PR #74, opened 2026-03-16
feat: Built-in agent — LLM-powered AEO analyst with chat API

arberx:
## Summary

- **Built-in LLM-powered agent** with chat API, thread persistence in SQLite, and 9 domain-specific AEO tools (visibility runs, timeline analysis, evidence lookup, competitor analysis, keyword management, sitemap inspection, etc.)
- **CLI command** (`canonry agent`) for interactive terminal chat sessions
- **Multi-provider support** — agent threads can use OpenAI, Claude, or Gemini as the backing LLM
- **Security hardening** — project ownership verification on thread endpoints, SSRF protection with DNS resolution on sitemap inspection, project-scoped run access
- **Bug fixes** — history windowing (newest messages kept, not oldest), complete evidence across all runs, malformed JSON recovery for Claude message replay

## Commits

1. `feat: built-in agent — LLM-powered AEO analyst with chat API` — core agent loop, tools, store, routes, CLI command, DB migrations
2. `fix(security): Add project ownership verification to thread endpoints`
3. `fix(agent): Add error handling for malformed JSON in tool call arguments`
4. `perf(agent): Move dynamic imports to top-level`
5. `refactor(agent): Replace circular HTTP self-calls with direct service layer`
6. `style(agent): Remove dead code and unused types`
7. `fix(agent): fix 5 bugs in agent loop, SSRF validation, and services` — history windowing, SSRF validation, project-scoped getRun, complete getHistory, Claude replay fix

## Test plan

- [x] CI passes (typecheck, lint, test)
- [x] SSRF validation tests cover IPv6 loopback, link-local, ULA, localhost, DNS resolution
- [ ] Manual: `canonry agent` interactive session against a project with runs
- [ ] Manual: verify thread persistence across server restarts

### DIFF
--- a/packages/api-routes/src/agent.ts
+++ b/packages/api-routes/src/agent.ts
@@ -184,7 +184,7 @@ export async function agentRoutes(app: FastifyInstance, opts: AgentRoutesOptions
       return reply.status(503).send({
         error: {
           code: 'AGENT_UNAVAILABLE',
-          message: 'Agent is not configured. Add a provider with an API key.',
+          message: 'Aero is not configured. Add a provider API key (claude, openai, or gemini) to enable the agent.',
         },
       })
     }
--- a/packages/api-routes/src/agent.ts
+++ b/packages/api-routes/src/agent.ts
@@ -155,7 +155,7 @@ export async function agentRoutes(app: FastifyInstance, opts: AgentRoutesOptions
       body: {
         type: 'object',
         properties: {
-          message: { type: 'string' },
+          message: { type: 'string', maxLength: 8000 },
         },
         required: ['message'],
       },
@@ -222,6 +222,7 @@ export async function agentRoutes(app: FastifyInstance, opts: AgentRoutesOptions
       return reply.status(404).send({ error: { code: 'NOT_FOUND', message: 'Thread not found' } })
     }
 
+    app.db.delete(agentMessages).where(eq(agentMessages.threadId, id)).run()
     app.db.delete(agentThreads).where(eq(agentThreads.id, id)).run()
 
     return reply.status(204).send()
--- a/packages/api-routes/src/agent.ts
+++ b/packages/api-routes/src/agent.ts
@@ -26,6 +26,9 @@ export interface AgentRoutesOptions {
 export async function agentRoutes(app: FastifyInstance, opts: AgentRoutesOptions) {
   const prefix = '/projects/:project/agent'
 
+  // Per-thread mutex — prevents concurrent agent loops from corrupting history
+  const activeThreads = new Set<string>()
+
   // ── Create thread ─────────────────────────────────────────
 
   app.post<{
@@ -41,8 +44,8 @@ export async function agentRoutes(app: FastifyInstance, opts: AgentRoutesOptions
       body: {
         type: 'object',
         properties: {
-          title: { type: 'string' },
-          channel: { type: 'string' },
+          title: { type: 'string', maxLength: 200 },
+          channel: { type: 'string', enum: ['chat', 'cli', 'api'] },
         },
       },
     },
@@ -186,9 +189,31 @@ export async function agentRoutes(app: FastifyInstance, opts: AgentRoutesOptions
       })
     }
 
-    const response = await opts.onAgentMessage(thread.projectId, threadId, message)
+    if (activeThreads.has(threadId)) {
+      return reply.status(409).send({
+        error: {
+          code: 'THREAD_BUSY',
+          message: 'This thread is already processing a message. Please wait.',
+        },
+      })
+    }
 
-    return reply.send({ threadId, response })
+    activeThreads.add(threadId)
+    try {
+      const response = await opts.onAgentMessage(thread.projectId, threadId, message)
+      return reply.send({ threadId, response })
+    } catch (err) {
+      const msg = err instanceof Error ? err.message : String(err)
+      const isLlmError = msg.includes('API error') || msg.includes('API key') || msg.includes('timeout')
+      return reply.status(isLlmError ? 502 : 500).send({
+        error: {
+          code: isLlmError ? 'LLM_ERROR' : 'AGENT_ERROR',
+          message: msg,
+        },
+      })
+    } finally {
+      activeThreads.delete(threadId)
+    }
   })
 
   // ── Delete thread ─────────────────────────────────────────
--- a/packages/api-routes/src/agent.ts
+++ b/packages/api-routes/src/agent.ts
@@ -20,6 +20,7 @@ export interface AgentRoutesOptions {
     projectId: string,
     threadId: string,
     message: string,
+    opts?: { provider?: string },
   ) => Promise<string>
 }
 
@@ -144,7 +145,7 @@ export async function agentRoutes(app: FastifyInstance, opts: AgentRoutesOptions
 
   app.post<{
     Params: { project: string; id: string }
-    Body: { message: string }
+    Body: { message: string; provider?: string }
   }>(`${prefix}/threads/:id/messages`, {
     schema: {
       params: {
@@ -159,13 +160,14 @@ export async function agentRoutes(app: FastifyInstance, opts: AgentRoutesOptions
         type: 'object',
         properties: {
           message: { type: 'string', maxLength: 8000 },
+          provider: { type: 'string', enum: ['openai', 'claude', 'gemini'] },
         },
         required: ['message'],
       },
     },
   }, async (request, reply) => {
     const { project, id: threadId } = request.params
-    const { message } = request.body
+    const { message, provider } = request.body
 
     const projectRow = resolveProject(app.db, project)
 
@@ -200,7 +202,7 @@ export async function agentRoutes(app: FastifyInstance, opts: AgentRoutesOptions
 
     activeThreads.add(threadId)
     try {
-      const response = await opts.onAgentMessage(thread.projectId, threadId, message)
+      const response = await opts.onAgentMessage(thread.projectId, threadId, message, { provider })
       return reply.send({ threadId, response })
     } catch (err) {
       const msg = err instanceof Error ? err.message : String(err)
--- a/packages/api-routes/src/agent.ts
+++ b/packages/api-routes/src/agent.ts
@@ -27,8 +27,10 @@ export interface AgentRoutesOptions {
 export async function agentRoutes(app: FastifyInstance, opts: AgentRoutesOptions) {
   const prefix = '/projects/:project/agent'
 
-  // Per-thread mutex — prevents concurrent agent loops from corrupting history
+  // Per-thread mutex — prevents concurrent agent loops from corrupting history.
+  // Also tracks pending work so it survives client disconnects.
   const activeThreads = new Set<string>()
+  const threadErrors = new Map<string, string>()
 
   // ── Create thread ─────────────────────────────────────────
 
@@ -138,7 +140,9 @@ export async function agentRoutes(app: FastifyInstance, opts: AgentRoutesOptions
       .orderBy(asc(agentMessages.createdAt))
       .all()
 
-    return reply.send({ ...thread, messages })
+    const processing = activeThreads.has(id)
+    const error = threadErrors.get(id) ?? null
+    return reply.send({ ...thread, messages, status: processing ? 'processing' : 'idle', error })
   })
 
   // ── Send message ──────────────────────────────────────────
@@ -201,21 +205,68 @@ export async function agentRoutes(app: FastifyInstance, opts: AgentRoutesOptions
     }
 
     activeThreads.add(threadId)
-    try {
-      const response = await opts.onAgentMessage(thread.projectId, threadId, message, { provider })
-      return reply.send({ threadId, response })
-    } catch (err) {
-      const msg = err instanceof Error ? err.message : String(err)
-      const isLlmError = msg.includes('API error') || msg.includes('API key') || msg.includes('timeout')
-      return reply.status(isLlmError ? 502 : 500).send({
-        error: {
-          code: isLlmError ? 'LLM_ERROR' : 'AGENT_ERROR',
-          message: msg,
-        },
+    threadErrors.delete(threadId)
+
+    // Fire-and-forget: the agent loop runs in the background so it
+    // survives client disconnects and page navigations.
+    opts.onAgentMessage(thread.projectId, threadId, message, { provider })
+      .catch((err) => {
+        const msg = err instanceof Error ? err.message : String(err)
+        threadErrors.set(threadId, msg)
+        app.log.error({ threadId, err: msg }, 'agent loop error')
       })
-    } finally {
-      activeThreads.delete(threadId)
+      .finally(() => {
+        activeThreads.delete(threadId)
+      })
+
+    return reply.status(202).send({ threadId, status: 'processing' })
+  })
+
+  // ── Update thread (rename) ───────────────────────────────
+
+  app.patch<{
+    Params: { project: string; id: string }
+    Body: { title?: string }
+  }>(`${prefix}/threads/:id`, {
+    schema: {
+      params: {
+        type: 'object',
+        properties: {
+          project: { type: 'string' },
+          id: { type: 'string' },
+        },
+        required: ['project', 'id'],
+      },
+      body: {
+        type: 'object',
+        properties: {
+          title: { type: 'string', maxLength: 200 },
+        },
+      },
+    },
+  }, async (request, reply) => {
+    const { project, id } = request.params
+    const { title } = request.body ?? {}
+
+    const projectRow = resolveProject(app.db, project)
+
+    const thread = app.db
+      .select()
+      .from(agentThreads)
+      .where(eq(agentThreads.id, id))
+      .get()
+
+    if (!thread || thread.projectId !== projectRow.id) {
+      return reply.status(404).send({ error: { code: 'NOT_FOUND', message: 'Thread not found' } })
     }
+
+    const updates: Record<string, unknown> = { updatedAt: new Date().toISOString() }
+    if (title !== undefined) updates.title = title
+
+    app.db.update(agentThreads).set(updates).where(eq(agentThreads.id, id)).run()
+
+    const updated = app.db.select().from(agentThreads).where(eq(agentThreads.id, id)).get()
+    return reply.send(updated)
   })
 
   // ── Delete thread ─────────────────────────────────────────
--- a/packages/api-routes/src/agent.ts
+++ b/packages/api-routes/src/agent.ts
@@ -1,0 +1,214 @@
+/**
+ * Agent API routes — chat with the built-in AEO analyst.
+ *
+ * POST /api/v1/projects/:project/agent/threads             — create thread
+ * GET  /api/v1/projects/:project/agent/threads              — list threads
+ * GET  /api/v1/projects/:project/agent/threads/:id          — get thread + messages
+ * POST /api/v1/projects/:project/agent/threads/:id/messages — send message
+ * DELETE /api/v1/projects/:project/agent/threads/:id        — delete thread
+ */
+
+import crypto from 'node:crypto'
+import { eq, desc, asc } from 'drizzle-orm'
+import type { FastifyInstance } from 'fastify'
+import { agentThreads, agentMessages } from '@ainyc/canonry-db'
+import { resolveProject } from './helpers.js'
+
+export interface AgentRoutesOptions {
+  /** Called when a user sends a message to the agent. Returns the agent's response. */
+  onAgentMessage?: (
+    projectId: string,
+    threadId: string,
+    message: string,
+  ) => Promise<string>
+}
+
+export async function agentRoutes(app: FastifyInstance, opts: AgentRoutesOptions) {
+  const prefix = '/projects/:project/agent'
+
+  // ── Create thread ─────────────────────────────────────────
+
+  app.post<{
+    Params: { project: string }
+    Body: { title?: string; channel?: string }
+  }>(`${prefix}/threads`, {
+    schema: {
+      params: {
+        type: 'object',
+        properties: { project: { type: 'string' } },
+        required: ['project'],
+      },
+      body: {
+        type: 'object',
+        properties: {
+          title: { type: 'string' },
+          channel: { type: 'string' },
+        },
+      },
+    },
+  }, async (request, reply) => {
+    const { project } = request.params
+    const { title, channel } = request.body ?? {}
+
+    const projectRow = resolveProject(app.db, project)
+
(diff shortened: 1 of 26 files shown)


packages/api-routes/src/agent.ts
```diff
@@ -0,0 +1,229 @@
...
+        properties: {
+          project: { type: 'string' },
+          id: { type: 'string' },
+        },
+        required: ['project', 'id'],
+      },
+      body: {
+        type: 'object',
+        properties: {
+          message: { type: 'string' },
```
arberx: **[Security]** No `maxLength` on the `message` field means a client can send an arbitrarily large string, causing the agent loop to pass it to the LLM and run up tokens/cost with a single request.

Suggested fix:
```typescript
message: { type: "string", maxLength: 8000 },
```
Ajv (used by Fastify) enforces this at schema validation time, so no extra code is needed in the handler.

packages/api-routes/src/agent.ts
```diff
@@ -0,0 +1,229 @@
...
+
+    const projectRow = resolveProject(app.db, project)
+
+    // Verify thread exists and belongs to this project
+    const thread = app.db
+      .select()
+      .from(agentThreads)
+      .where(eq(agentThreads.id, id))
+      .get()
+
```
arberx, line 299: **[Bug]** The DELETE handler removes the thread row but does not explicitly delete its `agent_messages` rows. The FK schema has `ON DELETE CASCADE`, so SQLite *should* handle this automatically — but only if SQLite foreign-key enforcement is enabled (`PRAGMA foreign_keys = ON`). If the DB connection does not set this pragma (common with drizzle-better-sqlite3 defaults), orphaned messages will accumulate silently.

Suggested fix: explicitly delete messages before deleting the thread, or verify the DB setup sets `foreign_keys = ON` and add a test that orphans are cleaned up:
```typescript
app.db.delete(agentMessages).where(eq(agentMessages.threadId, id)).run()
app.db.delete(agentThreads).where(eq(agentThreads.id, id)).run()
```

packages/canonry/src/server.ts
```diff
@@ -453,6 +458,72 @@ export async function createServer(opts: {
...
+  const llmConfig: LlmConfig = {
+    provider: llmProvider,
+    apiKey: registeredProvider.config.apiKey ?? '',
+    model: agentConf.model ?? registeredProvider.config.model,
+  }
+
+  const store = new AgentStore(db)
+  const services = new AgentServices(db)
+  const apiClient = new ApiClient(
+    opts.config.apiUrl,
```
arberx: **[Bug]** `registeredProvider.config.apiKey ?? ''` silently falls back to an empty string if the provider entry somehow lacks an API key. This means an agent handler is constructed and `buildAgentHandler` returns a function rather than `undefined` — so the route appears configured but every LLM call will get a 401. Better to guard early:
```typescript
if (!registeredProvider.config.apiKey) return undefined
```

packages/canonry/src/server.ts
```diff
@@ -453,6 +458,72 @@ export async function createServer(opts: {
...
+    opts.config.apiKey,
+  )
+
+  return async (projectId: string, threadId: string, message: string) => {
```
arberx: **[Bug]** `opts.config.apiUrl` and `opts.config.apiKey` are both optional in `CanonryConfig`. When the server is started without an external API key (the normal self-hosted path), `apiKey` is `undefined`, so every tool call routed through `ApiClient` (`run_sweep`, `get_gsc_*`, `inspect_url`) will fail at authentication.

The agent tools that depend on `ApiClient` should either use the internal service layer (as the other tools do) or be guarded with a check at startup:
```typescript
if (!opts.config.apiUrl) {
  // run_sweep and GSC tools will be unavailable
  app.log.warn('[agent] apiUrl not set; HTTP-backed agent tools will be unavailable')
}
```
Longer term, consider using the server's own base URL and an internal bearer token so self-hosted instances get full tool coverage.

packages/canonry/src/agent/loop.ts
```diff
@@ -0,0 +1,233 @@
...
+
+        // Find and execute tool
+        const tool = tools.find(t => t.name === toolName)
+        let result: string
+
+        if (tool) {
+          try {
+            result = await tool.execute(toolArgs)
+          } catch (err) {
+            result = `Error executing ${toolName}: ${err instanceof Error ? err.message : String(err)}`
```
arberx, line 260: **[Bug]** The assistant tool-call message is persisted *after* executing the tool (lines ~188–200), not before. If `tool.execute()` throws, the loop catches the error and adds a `tool` result message — but there is no corresponding `assistant` tool-call row in the DB. When the thread is later replayed via `getMessages`, the `tool` message will have a `toolCallId` but no matching `assistant` message with `tool_calls`, which will cause Claude's `convertToClaudeMessages` to produce an invalid message sequence and trigger an API error.

Move the `store.addMessage` for the assistant tool-call row to *before* `tool.execute()`, or include it in the same error branch.

packages/canonry/src/agent/loop.ts
```diff
@@ -0,0 +1,233 @@
...
+          toolArgs: null,
+          toolCallId: toolCall.id,
+        })
+
+        // Add tool result to conversation
+        messages.push({
+          role: 'tool',
+          content: result,
+          tool_call_id: toolCall.id,
+        })
```
arberx, line 280: **[Bug]** When `maxSteps` is reached, the loop returns the fallback message without persisting any in-flight tool-call messages that may have been added to `messages` but not yet stored. The fallback `store.addMessage` only stores the assistant text; a thread that hit the step limit can end up with an inconsistent DB state on replay. Consider flushing any un-persisted tool result messages before returning the fallback.

packages/canonry/src/agent/llm.ts
```diff
@@ -0,0 +1,248 @@
...
+  toolCalls?: ToolCall[]
+}
+
+const PROVIDER_ENDPOINTS: Record<string, string> = {
+  openai: 'https://api.openai.com/v1/chat/completions',
+  claude: 'https://api.anthropic.com/v1/messages',
+  gemini: 'https://generativelanguage.googleapis.com/v1beta/openai/chat/completions',
+}
+
+const DEFAULT_MODELS: Record<string, string> = {
```
arberx, line 46: **[Bug]** `PROVIDER_ENDPOINTS` has a `claude` entry (`https://api.anthropic.com/v1/messages`) but `chatCompletion` branches on `config.provider === 'claude'` and routes to `claudeCompletion` before ever reading `PROVIDER_ENDPOINTS`. The `claude` key in `PROVIDER_ENDPOINTS` is dead code and the URL is duplicated in `claudeCompletion` itself (line 155). This is currently harmless, but it will silently diverge if someone patches the endpoint map without touching `claudeCompletion`. Remove the `claude` entry from `PROVIDER_ENDPOINTS` and add a comment that Claude uses a dedicated code path.

packages/canonry/src/agent/services.ts
```diff
@@ -0,0 +1,153 @@
...
+      .where(inArray(querySnapshots.runId, runIds))
+      .all()
+
+    return {
+      project,
+      runs,
+      snapshots,
+    }
+  }
+
```
arberx, line 118: **[Performance]** `getTimeline` issues one `SELECT` per run to fetch its snapshots — classic N+1. With 50 runs that is 51 queries. Because SQLite is local/synchronous this is unlikely to be a bottleneck today, but the existing `getHistory` pattern already does a single bulk `inArray` fetch. Apply the same approach here:
```typescript
const runIds = runs.map(r => r.id)
const allSnapshots = runIds.length > 0
  ? this.db.select().from(querySnapshots).where(inArray(querySnapshots.runId, runIds)).all()
  : []

const snapshotsByRun = new Map<string, typeof allSnapshots>()
for (const s of allSnapshots) {
  const arr = snapshotsByRun.get(s.runId) ?? []
  arr.push(s)
  snapshotsByRun.set(s.runId, arr)
}

const timeline = runs.map(run => {
  const snapshots = snapshotsByRun.get(run.id) ?? []
  // ...
})
```

packages/canonry/src/agent/types.ts
```diff
@@ -0,0 +1,31 @@
...
+  createdAt: string
+  updatedAt: string
+}
+
+export interface AgentMessage {
+  id: string
+  threadId: string
+  role: 'user' | 'assistant' | 'tool'
+  content: string
+  toolName: string | null
```
arberx, line 19: **[Type Safety]** `AgentMessage.content` is typed as `string` (non-nullable), but assistant tool-call rows are stored with `content: 'Calling <toolName>'` as a synthetic placeholder. Meanwhile, the schema column is `TEXT NOT NULL` — which is consistent, but the semantic contract is misleading: `content` on a `role='assistant'` message with `toolName` set is a synthetic label, not actual LLM output. Consider either (a) making `content` nullable in the type and DB (`TEXT` with no `NOT NULL`) and storing `null` for tool-call rows, or (b) documenting clearly that content is a non-null sentinel string for tool-call rows, so future readers don't treat it as real assistant text.

packages/db/src/migrate.ts
```diff
@@ -218,6 +218,28 @@ const MIGRATIONS = [
...
+    created_at  TEXT NOT NULL,
+    updated_at  TEXT NOT NULL
+  )`,
+  `CREATE INDEX IF NOT EXISTS idx_agent_threads_project ON agent_threads(project_id)`,
+  `CREATE INDEX IF NOT EXISTS idx_agent_threads_updated ON agent_threads(updated_at)`,
+  `CREATE TABLE IF NOT EXISTS agent_messages (
+    id            TEXT PRIMARY KEY,
+    thread_id     TEXT NOT NULL REFERENCES agent_threads(id) ON DELETE CASCADE,
+    role          TEXT NOT NULL,
+    content       TEXT NOT NULL,
```
arberx, line 236: **[Type Safety]** The `role` column is `TEXT NOT NULL` with no `CHECK` constraint. Any string is valid at the DB level, so a code bug that stores e.g. `'system'` or a typo will silently corrupt history and cause subtle LLM API errors during replay. Add a constraint:
```sql
role TEXT NOT NULL CHECK(role IN ('user', 'assistant', 'tool'))
```
If adding a CHECK in the migration is too disruptive, at minimum add the constraint to the Drizzle schema definition so the generated types are correct.

packages/canonry/src/agent/loop.ts
```diff
@@ -0,0 +1,233 @@
+/**
```
arberx, line 1: **[Testing]** The new agent module (~830 lines across `loop.ts`, `store.ts`, `services.ts`, `llm.ts`, `tools.ts`) ships without a single unit test. The PR description marks manual tests as pending. Given the number of edge cases (tool-call replay, history windowing, malformed JSON recovery, maxSteps fallback, multi-provider routing), the risk of silent regressions is high.

At minimum, consider adding:
- `store.test.ts` — `addMessage` / `getMessages` windowing (newest-N-ascending)
- `loop.test.ts` — maxSteps fallback, JSON parse error in tool args, tool not found
- `services.test.ts` — `getTimeline` with zero runs, `getRun` project-scope check

These can use an in-memory SQLite DB the same way the existing `sitemap-parser.test.ts` spins up a local server.

packages/canonry/src/sitemap-parser.ts
```diff
@@ -19,24 +46,61 @@ function validateSitemapUrl(url: string): void {
...
+    if (isPrivateIP(host)) {
       throw new Error(`Sitemap URL points to a private or reserved IP range: ${url}`)
     }
+    return
+  }
+
+  // Resolve the hostname and verify all addresses are public
+  let addresses: string[]
+  try {
+    const results = await dns.resolve(host)
```
arberx, line 69: **[Security]** `dns.resolve6(host).catch(() => [] as string[])` silently ignores all errors, including real network failures. The concern is a TOCTOU window: `dns.resolve` (IPv4) succeeds with a public IP, the fetch proceeds, but between validation and fetch the DNS record changes. More concretely: if IPv4 resolution raises `ENODATA` (no A records) but the catch-all swallows it, and resolve6 also returns empty, `addresses` could be an empty array. The guard `if (addresses.length === 0)` covers that case, so this is not exploitable — but it could cause valid sitemaps with IPv6-only hosts to be spuriously rejected with "Cannot resolve sitemap hostname" instead of passing validation. Consider catching only `ENODATA`/`ENOTFOUND` specifically for `resolve6` and re-throwing other errors.

packages/canonry/src/server.ts
```diff
@@ -9,7 +9,8 @@ const { version: PKG_VERSION } = _require('../package.json') as { version: strin
 import Fastify from 'fastify'
 import type { FastifyInstance } from 'fastify'
 import { apiRoutes } from '@ainyc/canonry-api-routes'
```
arberx, line 11: **[Style]** `projects` is imported twice — once as `projects` and once aliased as `projectsTable` — from the same module on the same line. The alias is used in `buildAgentHandler` while the original name is used elsewhere. Clean this up by picking one canonical name for the import:
```typescript
import { auditLog, projects, type DatabaseClient } from '@ainyc/canonry-db'
```
and referencing `projects` consistently, or alias it at the `buildAgentHandler` call site instead.

arberx (commented):
## 🤖 Automated Review Summary

**Files reviewed:** 21  
**Comments left:** 13

**Issues found:**
- 🔴 Bug: 5
- 🟡 Security: 2
- 🟠 Performance: 1
- 🔵 Type Safety: 2
- 🟣 Testing: 1
- ⚪ Style: 1
- ⚪ Dead code: 1

### Key findings

**Bugs (fix before merge):**
1. **Tool-call persistence ordering** (`loop.ts` ~line 188) — assistant tool-call rows are stored *after* `tool.execute()` runs. If execution throws, the DB ends up with a `tool` result row but no matching `assistant` row, corrupting thread replay for Claude.
2. **Empty `apiKey` fallback** (`server.ts` ~line 499) — `apiKey ?? ''` means a misconfigured provider silently constructs a working handler that returns 401 on every LLM call instead of returning `undefined` to disable the agent.
3. **`ApiClient` with undefined `apiUrl/apiKey`** (`server.ts` ~line 503) — self-hosted instances without `apiUrl` set will get silent failures on `run_sweep` and all GSC tools.
4. **Orphaned messages on thread delete** (`agent.ts` ~line 220) — `ON DELETE CASCADE` only fires if `PRAGMA foreign_keys = ON`, which is not guaranteed.
5. **Unbounded `message` field** (`agent.ts` ~line 158) — no `maxLength` on the message body; a single large payload can rack up LLM token costs.

**Security:**
- `dns.resolve6` catch-all silently swallows errors (`sitemap-parser.ts`) — not exploitable but could incorrectly block IPv6-only hosts.
- Missing message length limit (covered above under Bugs).

**Testing gap:**
- ~830 lines of new agent code ship with zero unit tests. The loop's history-windowing, maxSteps fallback, and JSON recovery paths are all critical and unverified.

**Performance:**
- `getTimeline` has an N+1 query pattern; `getHistory` already demonstrates the correct bulk-fetch approach.

### What's done well ✅
- The SSRF hardening in `sitemap-parser.ts` is thorough: DNS resolution, IPv4/IPv6 loopback, link-local, ULA, and IPv4-mapped IPv6 addresses are all covered, with tests.
- History windowing (newest-N-ascending subquery) is the right approach and well-commented.
- Provider abstraction is clean — adding a new LLM is a one-function addition in `llm.ts`.
- Malformed JSON recovery in `convertToClaudeMessages` is a good defensive touch.
- DB schema with `ON DELETE CASCADE` + composite index on `(thread_id, created_at)` is solid.

**Overall assessment:** NEEDS_WORK on the tool-call persistence bug and the ApiClient/apiKey issues before this is safe to merge.

_This review was generated by an AI agent. Please verify all suggestions._

packages/api-routes/src/agent.ts
arberx: Fixed in 31bec0e.

packages/api-routes/src/agent.ts
arberx: Fixed in 31bec0e.

packages/canonry/src/server.ts
arberx: Fixed in 31bec0e.

packages/canonry/src/server.ts
arberx: Fixed in 31bec0e.

packages/canonry/src/agent/loop.ts
arberx: Fixed in 31bec0e.

packages/canonry/src/agent/loop.ts
arberx: Fixed in 31bec0e.

packages/canonry/src/agent/llm.ts
arberx: Fixed in 31bec0e.

packages/canonry/src/agent/services.ts
arberx: Fixed in 31bec0e.

packages/canonry/src/agent/types.ts
arberx: Fixed in 31bec0e.

packages/db/src/migrate.ts
arberx: Fixed in 31bec0e.

packages/canonry/src/agent/loop.ts
arberx: Fixed in 31bec0e.

packages/canonry/src/sitemap-parser.ts
arberx: Fixed in 31bec0e.

packages/canonry/src/server.ts
arberx: Fixed in 31bec0e.

packages/canonry/src/commands/agent.ts
```diff
@@ -0,0 +1,148 @@
...
+    if (opts?.format !== 'json') {
+      console.log(`Thread: ${threadId}\n`)
+    }
+  }
+
+  if (opts?.format !== 'json') {
+    process.stdout.write('Aero is thinking...')
+  }
+
+  await client.sendAgentMessage(project, threadId, message, opts?.provider)
```
arberx, line 58: **Missing error handling on `sendAgentMessage`:** If this call throws (e.g. network error, 4xx/5xx), the exception propagates uncaught and the polling loop below never runs — which is fine. However, if the server accepts the message but never transitions out of `'processing'`, the loop silently times out and returns an empty string (see line 62–80). A tighter pattern would be to `await` this, and then check the thread status before entering the poll loop.

packages/canonry/src/commands/agent.ts
```diff
@@ -0,0 +1,148 @@
...
+        console.error(`\nError: ${data.error}`)
+        process.exitCode = 1
+        return
+      }
+      break
+    }
+    if (opts?.format !== 'json') process.stdout.write('.')
+  }
+
+  if (opts?.format !== 'json') console.log('\n')
```
arberx, line 83: **Silent timeout — no error signal when the agent takes >3 minutes:** When `i` reaches 120 (120 × 1500ms = 3 min), the loop exits without setting `process.exitCode` or printing an error. The caller receives an empty `response` string and exit code 0, which looks like success.

```ts
// After the loop:
if (!response) {
  console.error('\nTimed out waiting for agent response (3 min).')
  process.exitCode = 1
  return
}
```

packages/canonry/src/sitemap-parser.ts
```diff
@@ -19,24 +46,66 @@ function validateSitemapUrl(url: string): void {
...
-async function parseSitemapRecursive(url: string, urls: Set<string>, depth: number): Promise<void> {
+async function parseSitemapRecursive(url: string, urls: Set<string>, depth: number, options?: FetchSitemapOptions): Promise<void> {
   if (depth > 3) return // Prevent infinite recursion
 
-  validateSitemapUrl(url)
+  if (!options?.dangerouslyAllowPrivate) {
+    await validateSitemapUrl(url)
+  }
 
   const res = await fetch(url)
```
arberx, line 110: **DNS rebinding / TOCTOU:** The DNS check in `validateSitemapUrl` (lines 66–88) happens before `fetch(url)`. A malicious DNS server can return a public IP during validation, then switch to a private IP for the actual fetch request — this is a classic DNS rebinding attack.

Full mitigation requires resolving the hostname to an IP, asserting it's public, then connecting to that specific IP directly (e.g. by passing a custom `agent` to `fetch` that pins the resolved IP). The current approach significantly raises the bar vs. the old static regex check, but it is not bulletproof. Worth adding a comment documenting this known limitation so it's not mistaken for a complete fix.

packages/canonry/src/server.ts
```diff
@@ -466,6 +471,90 @@ export async function createServer(opts: {
...
+  const registeredProvider = registry.get(defaultProvider as ProviderName)
+  if (!registeredProvider) return undefined
+
+  if (!registeredProvider.config.apiKey) return undefined
+
+  const store = new AgentStore(db)
+  const services = new AgentServices(db)
+
+  // ApiClient is only needed for HTTP-backed tools (run_sweep, GSC).
+  // If apiUrl/apiKey aren't set (self-hosted), those tools will gracefully error.
```
arberx, line 509: **Security: `systemTools` defaults to `false` — good. Consider adding a prominent warning when it's enabled.**

When `agent.systemTools: true`, the built-in agent gains shell execution, file I/O, and HTTP request capabilities. Since the agent operates on user-supplied message content, an adversarial input could abuse these tools. At minimum, log a startup warning when `systemTools` is `true` so operators notice, and document clearly in the config schema that this is a dangerous option.

arberx (commented):
🤖 **Automated Review — PR #74 (incremental: new commits since last review)**

**Summary:** Solid agent foundation. The LLM loop, tool system, and CLI commands are well-structured. A few issues flagged inline:

| Severity | Finding | File |
|----------|---------|------|
| 🟠 Bug | Silent timeout — loop exits with `exitCode=0` + empty response after 3 min | `commands/agent.ts:83` |
| 🟠 Bug | `--wait` polls `indexingState` (wrong field — means *allowed*, not *indexed*) | `commands/google.ts` (already in main) |
| 🔴 Security | DNS rebinding TOCTOU in `validateSitemapUrl` | `sitemap-parser.ts:110` |
| 🟡 Robustness | `sendAgentMessage` failure not linked to poll-loop lifecycle | `commands/agent.ts:58` |
| 🟡 Security | `systemTools: true` silently grants shell/file/network to agent | `server.ts:509` |

The `--wait` / `INDEXING_ALLOWED` bug is already on `main` — worth a hotfix or follow-up issue independent of this PR.

PR #108, opened 2026-03-18
refactor: TanStack Query + component extraction — modularize App.tsx monolith

arberx:
## Summary

Decompose the 6,665-line App.tsx monolith into modular architecture with TanStack Query for data fetching, caching, and smart refetch policies (phases 1-5 of planned 6-phase migration). Reduces App.tsx from 6,665 to 761 lines while extracting 43 new focused files.

## Architecture Changes

**Phase 1:** Helper extraction to `src/lib/` (base-path, tone, format, health, highlight utilities).  
**Phase 2:** Shared component extraction (Sparkline, ScoreGauge, ToneBadge, StatusBadge, etc.) to `src/components/shared/`.  
**Phase 3-4:** Layout and page decomposition (Sidebar, Topbar, Drawer, 7 page components, project sections, settings forms).  
**Phase 5:** TanStack Query data layer with `query-keys.ts`, `query-client.ts`, hooks (`use-dashboard`, `use-health`, `use-run-detail`), and mutation infrastructure with automatic cache invalidation.

## Test Coverage

All 15 existing app.test.tsx assertions pass with QueryClientProvider integration. Version bumped to 1.17.0.

## Next Phase (Deferred)

Phase 6 (TanStack Router) integrates type-safe routing and remains as separate future work.

### DIFF
--- a/apps/web/src/pages/ProjectPage.tsx
+++ b/apps/web/src/pages/ProjectPage.tsx
@@ -239,7 +239,10 @@ export function ProjectPage({
     a.href = url
     a.download = `${projectName}.yaml`
     a.click()
-    setTimeout(() => URL.revokeObjectURL(url), 100)
+    // Blob URL is revoked asynchronously — a.click() returns void with no
+    // completion signal, so revoking synchronously can break the download.
+    // The blob is small and will be GC'd when the page unloads.
+
   }
 
   async function handleAddKeywords() {


apps/web/src/queries/mutations.ts
```diff
@@ -0,0 +1,126 @@
...
+      void queryClient.invalidateQueries({ queryKey: queryKeys.projects.all })
+    },
+  })
+}
+
+export function useAddCompetitors() {
+  const queryClient = useQueryClient()
+  return useMutation({
+    mutationFn: async ({ projectName, domains }: { projectName: string; domains: string[] }) => {
+      const existing = await fetchCompetitors(projectName)
```
arberx, line 75: 🐛 **TOCTOU race condition in `useAddCompetitors`**

This is a read-modify-write pattern with no concurrency guard:
```ts
const existing = await fetchCompetitors(projectName)   // read
const merged = [...new Set([...existingDomains, ...domains])]
return setCompetitors(projectName, merged)             // write
```
If two calls overlap (e.g. user clicks "Add" twice quickly, or another tab is open), the second read may get stale data and the final write will silently drop the first call's additions.

Options:
1. Move the merge to the server and expose a `POST /projects/:name/competitors/append` endpoint (recommended — same issue exists in `useUpdateOwnedDomains`).
2. At minimum, also invalidate `queryKeys.projects.detail` so a stale `useAddCompetitors` in another hook can't cache the old list.
3. The UI already sets `competitorSaving` which disables the button — but this doesn't protect against concurrent sessions.

apps/web/src/pages/ProjectPage.tsx
```diff
@@ -0,0 +1,634 @@
...
+    return map
+  }, [locationTimeline])
+
+  const filteredEvidence = useMemo(() => {
+    const filtered = locationFilter !== undefined
+      ? model.visibilityEvidence.filter(e => locationFilter === '' ? !e.location : e.location === locationFilter)
+      : model.visibilityEvidence
+    if (!locationRunHistoryMap) return filtered
+    return filtered.map(item => {
+      const history = locationRunHistoryMap.get(`${item.keyword}::${item.provider}`)
```
arberx, line 187: 🐛 **`URL.revokeObjectURL` fires before the browser processes the download**

```ts
const url = URL.createObjectURL(blob)
const a = document.createElement('a')
a.href = url
a.download = `${model.project.name}.yaml`
a.click()
URL.revokeObjectURL(url)   // ← too early
```
`a.click()` is synchronous but the browser initiates the download asynchronously. Revoking immediately can produce a broken/empty download on some browsers (notably Firefox).

Fix:
```ts
a.click()
setTimeout(() => URL.revokeObjectURL(url), 100)
```

apps/web/src/pages/ProjectPage.tsx
```diff
@@ -0,0 +1,634 @@
...
+    { key: 'overview', label: 'Overview', href: `/projects/${model.project.id}` },
+    { key: 'search-console', label: 'Search Console', href: `/projects/${model.project.id}/search-console` },
+    { key: 'analytics', label: 'Analytics', href: `/projects/${model.project.id}/analytics` },
+  ]
+
+  return (
+    <div className="page-container">
+      {showDeleteConfirm ? (
+        <Card className="surface-card p-6 mb-6 border-rose-800/60">
+          <h3 className="text-base font-semibold text-rose-400 mb-2">Delete project?</h3>
```
arberx, line 321: 🐛 **`onDeleteProject` is called without `await` — `deleting` state is never reset**

```ts
onClick={async () => {
  setDeleting(true)
  onDeleteProject(model.project.name)   // ← fire-and-forget
}}
```
`onDeleteProject` is typed as `(projectName: string) => void` but callers likely return a Promise. As written, `setDeleting(true)` is set but never cleared — button stays permanently disabled if the deletion resolves or errors.

Fix:
```ts
onClick={async () => {
  setDeleting(true)
  try {
    await onDeleteProject(model.project.name)
  } finally {
    setDeleting(false)
  }
}}
```
Also update the prop type to `() => Promise<void>` to make the async contract explicit.

apps/web/src/queries/use-health.ts
```diff
@@ -0,0 +1,23 @@
+import { useQuery } from '@tanstack/react-query'
+import { fetchServiceStatus } from '../App.js'
+import type { HealthSnapshot, ServiceStatus } from '../view-models.js'
```
arberx, line 3: 🔒 **Circular import risk: `use-health.ts` imports from `App.js`**

```ts
import { fetchServiceStatus } from '../App.js'
```
`App.js` is the root component that almost certainly imports from the `queries/` folder. This creates a circular dependency: `App.js` → `use-health.ts` → `App.js`.

JavaScript bundlers can tolerate simple cycles, but module evaluation order can produce `undefined` at import time, causing silent runtime failures.

`fetchServiceStatus` belongs in `api.ts` (or a dedicated `api/health.ts`), not in the root component.

apps/web/src/components/project/GscSection.tsx
```diff
@@ -0,0 +1,1393 @@
...
+      setCoverage(null)
+      setCoverageHistory([])
+    } finally {
+      setLoadingCoverage(false)
+    }
+  }
+
+  async function handleRequestIndexing(urls: string[]) {
+    setRequestingIndexing(true)
+    setError(null)
```
arberx, line 167: 🐛 **`loadCoverage` silently swallows all errors**

```ts
} catch {
  setCoverage(null)
  setCoverageHistory([])
} finally {
  setLoadingCoverage(false)
}
```
Unlike `loadPerformanceRows` and `loadInspectionHistory`, this catch block never calls `setError(...)`. If `fetchGscCoverage` fails, the section renders empty with no feedback.

Add:
```ts
} catch (err) {
  setCoverage(null)
  setCoverageHistory([])
  setError(err instanceof Error ? err.message : 'Failed to load coverage data')
}
```

apps/web/src/components/project/ProjectSettingsSection.tsx
```diff
@@ -0,0 +1,371 @@
...
+      const loc: ApiLocation = { label, city, region, country: locCountry }
+      if (newLocTimezone.trim()) loc.timezone = newLocTimezone.trim()
+      await addLocation(project.name, loc)
+      await onUpdateProject(project.name, {})
+      setNewLocLabel('')
+      setNewLocCity('')
+      setNewLocRegion('')
+      setNewLocCountry('')
+      setNewLocTimezone('')
+      setShowAddLocation(false)
```
arberx, line 108: 🔒 **`onUpdateProject(project.name, {})` with empty updates used as a refetch trigger**

```ts
await addLocation(project.name, loc)
await onUpdateProject(project.name, {})   // used to trigger cache invalidation
```
Same pattern in `handleRemoveLocation` and `handleSetDefaultLocation`.

This calls `PATCH /projects/:name` with an empty body. If the server does a merge-patch, passing `{}` is safe; but if it treats missing keys as "clear this field," you'll silently wipe `displayName`, `canonicalDomain`, etc.

The location operations already mutate server state — the follow-up should be a targeted cache invalidation rather than a no-op update:
```ts
void queryClient.invalidateQueries({ queryKey: queryKeys.projects.all })
```
This requires passing `queryClient` down or exposing a `refetch` prop.

apps/web/src/components/settings/ProviderConfigForm.tsx
```diff
@@ -0,0 +1,170 @@
...
+    try {
+      const quota: { maxConcurrency?: number; maxRequestsPerMinute?: number; maxRequestsPerDay?: number } = {}
+      if (maxConcurrency.trim()) quota.maxConcurrency = parseInt(maxConcurrency.trim(), 10)
+      if (maxPerMinute.trim()) quota.maxRequestsPerMinute = parseInt(maxPerMinute.trim(), 10)
+      if (maxPerDay.trim()) quota.maxRequestsPerDay = parseInt(maxPerDay.trim(), 10)
+      await updateProviderConfig(providerName.toLowerCase(), {
+        ...(apiKey.trim() ? { apiKey: apiKey.trim() } : {}),
+        ...(baseUrl.trim() ? { baseUrl: baseUrl.trim() } : {}),
+        ...(model.trim() ? { model: model.trim() } : {}),
+        ...(Object.keys(quota).length > 0 ? { quota } : {}),
```
arberx, line 60: 🧩 **`parseInt` on unvalidated user input — `NaN` will be sent to the API**

```ts
if (maxConcurrency.trim()) quota.maxConcurrency = parseInt(maxConcurrency.trim(), 10)
```
`parseInt('abc', 10)` returns `NaN`. The `if (x.trim())` guard only rejects empty strings — `"abc"` or `"1e5"` pass through. `NaN` serializes to `null` in JSON, which the API likely rejects or silently ignores.

Fix:
```ts
const parseQuotaField = (s: string): number | undefined => {
  const n = parseInt(s.trim(), 10)
  return Number.isFinite(n) && n > 0 ? n : undefined
}
const maxConcurrencyVal = parseQuotaField(maxConcurrency)
if (maxConcurrencyVal !== undefined) quota.maxConcurrency = maxConcurrencyVal
```

arberx (commented):
## 🤖 Automated Review Summary

Reviewed 50 changed files (~13.8k lines). Found **4 bugs**, **2 security/correctness concerns**, **2 type-safety issues**, and **missing test coverage** for new utilities and hooks. Overall the refactor is clean — main things to address before merge are the race condition in `useAddCompetitors`, the `onDeleteProject` await bug, and the `parseInt`/`NaN` issue.

**Legend:** 🐛 Bug · 🔒 Security/Correctness · 🧩 Type Safety · ✅ Missing Tests

apps/web/src/queries/mutations.ts
```diff
@@ -0,0 +1,126 @@
...
+      updates: {
+        displayName?: string
+        canonicalDomain?: string
+        ownedDomains?: string[]
+        country?: string
+        language?: string
+      }
+    }) => updateProject(projectName, updates),
+    onSuccess: () => {
+      void queryClient.invalidateQueries({ queryKey: queryKeys.projects.all })
```
arberx, line 115: 🧩 **Fragile conditional type for `useCreateProject` mutationFn parameter**

```ts
mutationFn: ({ name, body }: Parameters<typeof createProject> extends [infer N, infer B]
  ? { name: N; body: B }
  : never) => createProject(name, body),
```
If `createProject`'s signature changes (or inference fails), this resolves to `never` with no clear error pointing here.

Prefer an explicit type:
```ts
mutationFn: ({ name, body }: { name: string; body: CreateProjectBody }) =>
  createProject(name, body),
```
(export `CreateProjectBody` from `api.ts`). Clearer errors, easier to refactor.

apps/web/src/pages/ProjectPage.tsx
```diff
@@ -0,0 +1,634 @@
...
+  onOpenEvidence,
+  onOpenRun,
+  onTriggerRun,
+  onDeleteProject,
+  onAddKeywords,
+  onDeleteKeywords: _onDeleteKeywords,
+  onAddCompetitors,
+  onUpdateOwnedDomains,
+  onUpdateProject,
+  onNavigate,
```
arberx: ✅ **`onDeleteKeywords` is destructured but never used**

```ts
onDeleteKeywords: _onDeleteKeywords,
```
If keyword deletion is not yet implemented, add a `// TODO` comment so it doesn't get lost. If the feature is genuinely removed, drop the prop from the interface to keep the API surface clean.

apps/web/src/queries/use-dashboard.ts
```diff
@@ -0,0 +1,109 @@
+import { useCallback, useMemo } from 'react'
```
arberx, line 1: ✅ **Missing tests for `useDashboard` hook**

This is the central data-fetching hook with non-trivial logic: conditional `enabled`, dynamic `useQueries` fan-out, `allProjectDetailsLoaded` gate, polling interval derived from run status. None of this is covered — `queries.test.ts` only exercises `queryKeys`.

Recommend adding `renderHook` tests (with a mock `QueryClient`) for:
1. Returns `null` dashboard while project details are loading
2. Returns `initialDashboard` immediately when provided (no fetches fired)
3. `refetchInterval` returns `3000` when a run is `running` or `queued`

apps/web/src/lib/format-helpers.ts
```diff
@@ -0,0 +1,90 @@
+export function formatErrorLog(error: string): string {
```
arberx, line 1: ✅ **`parsePreset` / `scheduleLabel` / `buildPreset` are pure functions with no test coverage**

These have multiple branches (daily, twice-daily, weekly@{day}, custom, fallback) and are easy to break silently. Suggest adding a `format-helpers.test.ts`:
```ts
it('parsePreset handles daily@9', () =>
  expect(parsePreset('daily@9', '')).toEqual({ freq: 'daily', hour: 9, customCron: '' }))
it('parsePreset falls back to custom for unknown preset', () =>
  expect(parsePreset('bogus', '0 5 * * *')).toEqual({ freq: 'custom', hour: 6, customCron: '0 5 * * *' }))
it('scheduleLabel returns custom cron when preset is null', () =>
  expect(scheduleLabel(null, '0 5 * * *', 'UTC')).toBe('Custom: 0 5 * * * · UTC'))
```

apps/web/src/lib/health-helpers.ts
```diff
@@ -0,0 +1,104 @@
+import type { HealthSnapshot, MetricTone, SettingsVm, SetupWizardVm, SystemHealthCardVm } from '../view-models.js'
```
arberx, line 1: ✅ **`getLaunchBlockedReason` and `buildSetupModel` have no test coverage**

`getLaunchBlockedReason` gates whether the setup wizard's launch button is enabled — a critical user flow with 4 distinct blocking conditions. A regression here would silently allow launching in a broken state (or permanently block a healthy one). Since these are pure functions, they're trivial to unit-test.

PR #114, opened 2026-03-18
fix: base-path-aware API route prefix and SPA fallback

### LINKED ISSUE #113: bug: --base-path breaks API routes — SPA catch-all returns HTML for all /canonry/api/v1/* requests
## Summary

When `canonry serve --base-path /canonry/` is used, all API calls return `200 OK` with the SPA's `index.html` instead of JSON. The API is completely unreachable.

## Root Cause

Three separate issues in `packages/canonry/src/server.ts`:

### 1. `setNotFoundHandler` API check is not base-path aware

```ts
// line ~504 — current code
if (request.url.startsWith('/api/')) {
  return reply.status(404).send({ error: 'Not found', path: request.url })
}
```

With `--base-path /canonry/`, API routes are at `/canonry/api/v1/...` — not `/api/...`. The guard never matches, so unmatched routes return HTML instead of JSON 404.

### 2. API routes are registered without base-path prefix

```ts
// packages/api-routes/src/index.ts line ~168
}, { prefix: '/api/v1' })
```

The prefix is hardcoded. When Fastify receives `/canonry/api/v1/projects`, no route matches → falls through to the SPA catch-all.

### 3. `@fastify/static` and root handler don't account for base-path

Static assets are served at prefix `/` and the root route is `app.get('/')` — regardless of the configured base path.

## Reproduction

```bash
canonry serve --port 4100 --base-path /canonry/

# Returns SPA HTML (200) instead of JSON
curl http://localhost:4100/canonry/api/v1/health
curl http://localhost:4100/canonry/api/v1/projects
```

Without `--base-path` everything works correctly.

## Note on Caddy workaround

In the current production setup, Caddy strips the `/canonry/` prefix before forwarding to canonry, so the server receives `/api/v1/...` and works correctly. But running canonry directly (e.g. for local dev, or without a reverse proxy) is broken when `--base-path` is set.

## Proposed Fix

Pass `basePath` into the server and use it in three places:

```ts
// 1. Register API routes under base-path prefix
const apiPrefix = basePath ? `${basePath}api/v1` : '/api/v1'
await app.register(apiRoutes, { ..., prefix: apiPrefix })

// 2. Fix the not-found handler guard
const apiPathPrefix = basePath ? `${basePath}api/` : '/api/'
if (request.url.startsWith(apiPathPrefix) || request.url.startsWith('/api/')) {
  return reply.status(404).send({ error: 'Not found', path: request.url })
}

// 3. Register static assets and root route under base-path
await app.register(fastifyStatic.default, {
  root: assetsDir,
  prefix: basePath ?? '/',
  ...
})
app.get(basePath ?? '/', handler)
```

## Context

Discovered while testing v1.19.3 after merging PRs #107–#112. The Caddy-based production setup is unaffected, but direct invocation with `--base-path` is broken.

arberx:
Fixes #113.

## Problem

When `canonry serve --base-path /canonry/` is used, the server is completely non-functional:
- All API calls to `/canonry/api/v1/*` return `200 OK` with the SPA's `index.html` instead of JSON
- The root `/` and static assets aren't served under the base path
- The `setNotFoundHandler` API guard only checked `/api/` prefix, missing base-path-prefixed routes

The Caddy reverse proxy masked this in production (Caddy strips `/canonry/` before forwarding), so it was only caught when running canonry directly with `--base-path`.

## Root Causes

Three issues in `server.ts`:

1. **API routes** registered at hardcoded `/api/v1` regardless of base path
2. **`setNotFoundHandler`** guard: `request.url.startsWith('/api/')` — missed `/canonry/api/v1/*`
3. **Static plugin + root handler** registered at `/` regardless of base path

## Changes

**`packages/api-routes/src/index.ts`**
- Added `routePrefix?: string` to `ApiRoutesOptions` (named `routePrefix` not `prefix` to avoid collision with Fastify's reserved `prefix` register option — using `prefix` would double-prefix all routes)
- Use `opts.routePrefix ?? '/api/v1'` in the inner register call

**`packages/canonry/src/server.ts`**
- Compute `basePath` once before `apiRoutes` registration (moved out of the static block)
- Pass `routePrefix: apiPrefix` to `apiRoutes` where `apiPrefix = basePath ? `${basePath}api/v1` : '/api/v1'`
- Register `@fastify/static` and root handler at `basePath ?? '/'`
- Update `setNotFoundHandler` guard to check both `/api/` and `${basePath}api/`
- Register `/health` at both `/health` and `${basePath}health` when base path is set

## Test Results

- `pnpm typecheck` ✅
- `pnpm test` ✅ (497 tests, 0 failures)

Manual verification with `--base-path /canonry/`:
```
GET /canonry/api/v1/health     → 200 JSON  ✅  (was: 200 HTML)
GET /canonry/api/v1/projects   → 200 JSON  ✅  (was: 200 HTML)
GET /canonry/health            → 200 JSON  ✅
GET /canonry/api/v1/missing    → 404 JSON  ✅  (was: 200 HTML)
GET /canonry/                  → 200 HTML  ✅
```

### DIFF
--- a/packages/api-routes/src/index.ts
+++ b/packages/api-routes/src/index.ts
@@ -39,7 +39,8 @@ export interface ApiRoutesOptions {
   /** Skip auth for testing */
   skipAuth?: boolean
   /** API route prefix (default: /api/v1). Override for sub-path deployments e.g. /canonry/api/v1.
-   *  Named routePrefix (not prefix) to avoid collision with Fastify's reserved prefix option. */
+   *  Named routePrefix (not prefix) to avoid collision with Fastify's reserved prefix option.
+   *  Must start with '/' — values without a leading slash will be rejected at startup. */
   routePrefix?: string
   /** Callback when a run is created (wire up job runner) */
   onRunCreated?: (runId: string, projectId: string, providers?: string[], location?: import('@ainyc/canonry-contracts').LocationContext | null) => void
@@ -75,6 +76,14 @@ export interface ApiRoutesOptions {
 }
 
 export async function apiRoutes(app: FastifyInstance, opts: ApiRoutesOptions) {
+  // Validate routePrefix format eagerly to surface misconfiguration at startup
+  // rather than silently mis-routing all API requests.
+  if (opts.routePrefix !== undefined && !opts.routePrefix.startsWith('/')) {
+    throw new Error(
+      `apiRoutes: routePrefix must start with '/' — got ${JSON.stringify(opts.routePrefix)}`,
+    )
+  }
+
   // Decorate with db
   app.decorate('db', opts.db)
 
--- a/packages/canonry/src/server.ts
+++ b/packages/canonry/src/server.ts
@@ -220,10 +220,14 @@ export async function createServer(opts: {
 
   // Resolve base path early so API route prefix and SPA handler both use it.
   // Normalize: ensure it starts and ends with '/' (e.g. '/canonry/').
+  // A value that normalises to bare '/' is treated as no base path to avoid
+  // a duplicate-route error with fastify-static (which also registers '/').
   const rawBasePath = process.env.CANONRY_BASE_PATH ?? opts.config.basePath
-  const basePath: string | undefined = rawBasePath
+  const normalizedBasePath = rawBasePath
     ? ('/' + rawBasePath.replace(/^\//, '').replace(/\/?$/, '/'))
     : undefined
+  const basePath: string | undefined =
+    normalizedBasePath === '/' ? undefined : normalizedBasePath
 
   // API routes are registered under <basePath>api/v1 (or /api/v1 when no base path).
   const apiPrefix = basePath ? `${basePath}api/v1` : '/api/v1'
@@ -494,25 +498,43 @@ export async function createServer(opts: {
       index: false,
     })
 
-    // Serve index.html with injected config for the root/base-path route
-    const rootRoute = basePath ?? '/'
-    app.get(rootRoute, (_request, reply) => {
+    // Serve index.html with injected config for the root/base-path route.
+    // Register both the trailing-slash form ('/canonry/') and the bare form
+    // ('/canonry') so either URL shape hits the handler without a 404.
+    const serveIndex = (_request: unknown, reply: { type: (t: string) => { send: (b: string) => unknown }; status: (s: number) => { send: (b: unknown) => unknown } }) => {
       if (fs.existsSync(indexPath)) {
         const html = fs.readFileSync(indexPath, 'utf-8')
         return reply.type('text/html').send(injectConfig(html))
       }
       return reply.status(404).send({ error: 'Dashboard not built' })
-    })
+    }
+    const rootRouteTrailing = basePath ?? '/'
+    app.get(rootRouteTrailing, serveIndex)
+    // Also register the no-trailing-slash variant when base path is set
+    // (e.g. '/canonry' in addition to '/canonry/') to avoid a 404 on bare navigation.
+    if (basePath) {
+      const rootRouteBare = basePath.replace(/\/$/, '')
+      if (rootRouteBare) app.get(rootRouteBare, serveIndex)
+    }
 
-    // SPA fallback: serve index.html for unmatched non-API routes
+    // SPA fallback: serve index.html for unmatched routes that belong to this app.
+    // - With no base path: serve for any non-API path (existing behaviour).
+    // - With base path: only serve for paths under basePath to avoid hijacking
+    //   other apps co-hosted on the same origin outside the base path.
     app.setNotFoundHandler((request, reply) => {
+      const url = request.url.split('?')[0]!
+
       // Never serve HTML for API routes — return proper JSON 404.
-      // Check both the prefixed path (/canonry/api/...) and the bare path (/api/...)
-      // so this guard works with and without --base-path.
+      // Guard against both the prefixed (/canonry/api/...) and bare (/api/...) forms.
       const isApiRoute =
-        request.url.startsWith('/api/') ||
-        (basePath !== undefined && request.url.startsWith(`${basePath}api/`))
+        url.startsWith('/api/') ||
+        (basePath !== undefined && url.startsWith(`${basePath}api/`))
       if (isApiRoute) {
+        return reply.status(404).send({ error: 'Not found', path: request.url })
+      }
+
+      // When a base path is configured, only serve the SPA for paths under it.
+      if (basePath && !url.startsWith(basePath)) {
         return reply.status(404).send({ error: 'Not found', path: request.url })
       }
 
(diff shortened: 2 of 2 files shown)


packages/canonry/src/server.ts
```diff
@@ -482,15 +487,16 @@ export async function createServer(opts: {
...
       wildcard: false,
       // Don't serve index.html automatically — we handle it with config injection
       serve: true,
       index: false,
     })
 
-    // Serve index.html with injected API key for the root route
-    app.get('/', (_request, reply) => {
+    // Serve index.html with injected config for the root/base-path route
+    const rootRoute = basePath ?? '/'
```
arberx: **[Bug]** `rootRoute` will be `/canonry/` (trailing slash) when `basePath` is set, but Fastify does **not** treat `/canonry/` and `/canonry` as the same route by default (`ignoreTrailingSlash` defaults to `false`). Users visiting `/canonry` without the trailing slash will get a 404 from the router rather than the SPA.

Suggested fix — strip the trailing slash for the route handler, keeping the normalised `basePath` for other uses:
```typescript
const rootRoute = basePath ? basePath.replace(/\/$/, ) : /
```
Alternatively, also register a redirect handler:
```typescript
if (basePath) {
  const withoutSlash = basePath.replace(/\/$/, )
  app.get(withoutSlash, (_req, reply) => reply.redirect(301, basePath!))
}
```

packages/canonry/src/server.ts
```diff
@@ -500,8 +506,13 @@ export async function createServer(opts: {
 
     // SPA fallback: serve index.html for unmatched non-API routes
     app.setNotFoundHandler((request, reply) => {
-      // Never serve HTML for API routes — return proper JSON 404
-      if (request.url.startsWith('/api/')) {
+      // Never serve HTML for API routes — return proper JSON 404.
+      // Check both the prefixed path (/canonry/api/...) and the bare path (/api/...)
+      // so this guard works with and without --base-path.
+      const isApiRoute =
```
arberx, line 529: **[Bug]** When `basePath` is set, this `setNotFoundHandler` will serve `index.html` for **any** unmatched path on the server (e.g. `/random-page`, `/other-app/foo`), not just paths under `basePath`. With sub-path deployments the server is more likely to be reverse-proxied alongside other apps, making the over-broad SPA fallback a real problem.

Suggested fix — gate the SPA fallback on the request being under `basePath`:
```typescript
if (basePath && !request.url.startsWith(basePath)) {
  return reply.status(404).send({ error: "Not found", path: request.url })
}
```
Add this check immediately after the `isApiRoute` guard.

packages/canonry/src/server.ts
```diff
@@ -513,12 +524,19 @@ export async function createServer(opts: {
...
 
-  // Health endpoint
+  // Health endpoint — registered at both /health and <basePath>health when base path is set
   app.get('/health', async () => ({
     status: 'ok',
     service: 'canonry',
     version: PKG_VERSION,
   }))
+  if (basePath) {
+    app.get(`${basePath}health`, async () => ({
```
arberx: **[Style]** The health endpoint handler body is copy-pasted verbatim. Extract it to avoid drift if the response shape ever changes:

```typescript
const healthHandler = async () => ({ status: "ok", service: "canonry", version: PKG_VERSION })
app.get("/health", healthHandler)
if (basePath) {
  app.get(`${basePath}health`, healthHandler)
}
```

packages/api-routes/src/index.ts
```diff
@@ -38,6 +38,9 @@ export interface ApiRoutesOptions {
   openApiInfo?: OpenApiInfo
   /** Skip auth for testing */
   skipAuth?: boolean
+  /** API route prefix (default: /api/v1). Override for sub-path deployments e.g. /canonry/api/v1.
+   *  Named routePrefix (not prefix) to avoid collision with Fastify's reserved prefix option. */
+  routePrefix?: string
```
arberx, line 44: **[Bug]** `routePrefix` is consumed directly as Fastify's `prefix` with no format validation. If a caller passes a value without a leading `/` (e.g. `"api/v1"`) Fastify will silently mis-route all API calls. Likewise, the `isApiRoute` guard in `server.ts` assumes the prefix starts with the base path — a malformed value would bypass that guard.

Consider adding a startup assertion:
```typescript
if (routePrefix && !routePrefix.startsWith("/")) {
  throw new Error(`routePrefix must start with "/", got: ${routePrefix}`)
}
```

packages/canonry/src/server.ts
```diff
@@ -218,9 +218,20 @@ export async function createServer(opts: {
     },
   } as const
 
+  // Resolve base path early so API route prefix and SPA handler both use it.
+  // Normalize: ensure it starts and ends with '/' (e.g. '/canonry/').
+  const rawBasePath = process.env.CANONRY_BASE_PATH ?? opts.config.basePath
+  const basePath: string | undefined = rawBasePath
+    ? ('/' + rawBasePath.replace(/^\//, '').replace(/\/?$/, '/'))
```
arberx, line 227: **[Bug]** When `rawBasePath` is already `"/"` (i.e. the root), the normalization produces `"/"` as `basePath`, making `apiPrefix` equal to `"/api/v1"` and `rootRoute` equal to `"/"` — which is correct. However `fastify-static` then gets `prefix: "/"` and `app.get("/", ...)` is registered in *addition* to whatever the static plugin already serves, which could cause a route conflict in Fastify (`FST_ERR_DUP_ROUTE`).

Consider treating `rawBasePath === "/"` the same as `undefined` (no sub-path) so both branches behave identically:
```typescript
const basePath: string | undefined = rawBasePath && rawBasePath !== "/"
  ? ("/" + rawBasePath.replace(/^\//, "").replace(/\/?$/, "/"))
  : undefined
```

arberx (commented):
## 🤖 Automated Review Summary

**PR:** fix: base-path-aware API route prefix and SPA fallback
**Files reviewed:** `packages/api-routes/src/index.ts`, `packages/canonry/src/server.ts`

### ✅ What looks good
- Moving base-path resolution to the top of `createServer` is the right call — it eliminates the previous duplication and makes it impossible for the two consumers (API prefix + SPA handler) to diverge.
- Using `routePrefix` instead of `prefix` in `ApiRoutesOptions` cleanly avoids an `opts.prefix` name shadowing the Fastify plugin-registration option.
- The `isApiRoute` guard checking both `/api/` and `${basePath}api/` is a sensible safety net.
- Registering both `/health` and `${basePath}health` ensures load-balancer probes work regardless of whether the proxy strips the prefix.

### 🐛 Issues found (4 inline comments)

| # | Severity | Location | Issue |
|---|----------|----------|-------|
| 1 | 🔴 Bug | `server.ts:498` | `rootRoute = "/canonry/"` — Fastify treats `/canonry/` and `/canonry` as distinct routes; bare `/canonry` returns 404 |
| 2 | 🔴 Bug | `server.ts:512` | SPA `setNotFoundHandler` serves `index.html` for **all** unmatched paths, including those outside `basePath`; breaks co-hosted apps |
| 3 | 🟡 Bug | `server.ts:225` | `rawBasePath = "/"` normalises to `"/"` and takes the `basePath` truthy branch, potentially causing a duplicate-route error with fastify-static |
| 4 | 🟡 Bug | `api-routes/index.ts:43` | `routePrefix` is passed to Fastify with no format validation; a value without a leading `/` will silently mis-route all API requests |
| 5 | 🔵 Style | `server.ts:534` | Health handler body is copy-pasted; extract to a shared constant to avoid drift |

### 🔑 Priority fix
Issue #1 (trailing-slash mismatch on `rootRoute`) will cause a user-visible 404 for every browser that navigates to `/canonry` instead of `/canonry/`. Recommend fixing before merge.

PR #330, opened 2026-04-17
fix(db): extract legacy credentials before dropping columns

arberx:
## Summary

Fixes the silent data-loss risk in PR #326's column drop: `migrate()` previously dropped legacy credential columns (`access_token`, `refresh_token`, `token_expires_at`, `private_key`) before `createServer()` had a chance to migrate them to `~/.canonry/config.yaml`, so users upgrading directly to a version with the drop would lose their stored Google OAuth tokens and GA4 service-account keys.

The fix splits the migration into two steps in `packages/db/src/migrate.ts`: a new `extractAndDropLegacyCredentials(db)` SELECTs rows from the legacy columns, then drops each column atomically (idempotent via `pragma_table_info` checks). `server.ts` calls this once at startup and applies the extracted rows to config via `applyLegacyCredentials()`. Tests cover migration of OAuth tokens and GA4 keys, the skip-when-already-present case, idempotency, and config.yaml preservation.

Also bundles small cleanups carried over on this branch (provider normalize tweaks, agent-bootstrap simplification, evidence modal highlight fix) and bumps to 1.48.3.

## Test plan
- [x] `pnpm typecheck`
- [x] `pnpm test` (991 passed)
- [x] `pnpm lint`


packages/db/src/migrate.ts
```diff
@@ -453,6 +458,142 @@ function isDuplicateColumnError(err: unknown): boolean {
...
+      access_token: string | null
+      refresh_token: string
+      token_expires_at: string | null
+      scopes: string
+      created_at: string
+      updated_at: string
+    }>
+    for (const row of rows) {
+      let scopes: string[] = []
+      try { scopes = JSON.parse(row.scopes || '[]') as string[] } catch { scopes = [] }
```
arberx: **JSON Column Parsing rule violation.**

This is the exact anti-pattern explicitly forbidden by the [root `AGENTS.md` “JSON Column Parsing (Critical)” section](https://github.com/AINYC/canonry/blob/main/AGENTS.md#json-column-parsing-critical):

> Use `parseJsonColumn(value, fallback)` from `@ainyc/canonry-db`. Never write `JSON.parse(row.field || '[]') as SomeType[]` — this pattern is fragile (missing fallback = crash, wrong cast = silent corruption).

`row.scopes` is a raw SQLite text column read via `db.all(sql.raw(...))` on `google_connections`, so the rule applies. `parseJsonColumn` is defined in `packages/db/src/json.ts` and re-exported from `packages/db/src/index.ts`, so it’s importable locally from `migrate.ts`. Replace with `parseJsonColumn<string[]>(row.scopes, [])`.

packages/canonry/src/server.ts
```diff
@@ -289,8 +244,11 @@ export async function createServer(opts: {
     }
   }
 
-  // Migrate credentials from DB to config.yaml (one-time upgrade for pre-1.45.1 installs)
-  migrateDbCredentialsToConfig(opts.db, opts.config)
+  // One-time upgrade for pre-1.45.1 installs: extract any credentials still
+  // stored in the DB, drop the legacy columns, and persist the values into
+  // config.yaml. Order matters — extraction must run before any code path
+  // that depends on the columns being absent.
+  applyLegacyCredentials(extractAndDropLegacyCredentials(opts.db), opts.config)
```
arberx: **Data-loss regression: credentials are permanently gone if `saveConfigPatch` fails after columns are already dropped.**

`extractAndDropLegacyCredentials` reads the rows AND runs `ALTER TABLE ... DROP COLUMN` in the same call (see `packages/db/src/migrate.ts:562-564` and `:591`). SQLite's DROP COLUMN rewrites the table — there is no retention. `applyLegacyCredentials` then calls `saveConfigPatch({ google: ... })` at server.ts:184 and `saveConfigPatch({ ga4: ... })` at server.ts:203, both of which invoke `fs.writeFileSync` and can throw on EACCES / ENOSPC / EROFS / YAML serialization errors.

If either `saveConfigPatch` throws, the rows were already dropped from the DB but never persisted to `~/.canonry/config.yaml` — the in-memory rows unwind with the stack and the credentials are permanently lost. The prior `migrateDbCredentialsToConfig` (removed here) never dropped columns, so a config-write failure was retried on next boot. A partial-failure variant also exists: a throw during the Google `saveConfigPatch` leaves GA4 extracted (columns dropped) but unwritten too.

Fix: write to config first (and confirm the write succeeded) before dropping the columns — e.g. split `extractAndDropLegacyCredentials` into `extractLegacyCredentials` + a separate `dropLegacyCredentialColumns` the caller invokes only after `saveConfigPatch` returns cleanly.

arberx: **Regression: legacy migration is no longer best-effort; unexpected errors now crash server startup.**

The prior `migrateDbCredentialsToConfig` wrapped its entire body in `try { ... } catch {}` with the explicit comment “Migration is best-effort — if the columns don't exist or the query fails, the user can re-authenticate. Don't block server startup.” The new call site has no outer try/catch, and `extractAndDropLegacyCredentials` only swallows `no such column: <name>` errors inside `dropColumnIfExists` (see `packages/db/src/migrate.ts:496-512`).

Every other failure path now propagates out of `createServer()`: a failing `pragma_table_info` probe, `SELECT` errors on a locked/corrupt DB, `INNER JOIN projects` failures on `ga_connections`, and any filesystem error from `saveConfigPatch` (read-only `~/.canonry`, full disk, unwritable config file). Users who previously could boot and re-authenticate will be stuck in a startup crash loop. If the intent is to no longer be best-effort, that’s a deliberate trade-off worth calling out in the PR description; otherwise, restore the outer try/catch at the call site or inside `applyLegacyCredentials`.

arberx (commented):
Automated code review: 3 validated issues on the legacy credential migration path.

PR #332, opened 2026-04-18
Native Aero agent loop: rip out OpenClaw, ship built-in agent with dashboard bar

arberx:
## Summary

Rips out the OpenClaw-backed agent runtime and ships **Aero**, a built-in AI agent powered by [`@mariozechner/pi-agent-core`](https://github.com/badlogic/pi-mono). Adds a CLI (`canonry agent ask`), a dashboard command bar, proactive wake-up on run completion, and 13 typed tools.

Users with their own agent (Claude Code, Codex, custom) drive canonry through the regular CLI/API — same surface as any human operator. `canonry agent attach --url …` is a **separate** concern: it registers a push-notification webhook for agents that want to be woken up by `run.completed` / insight events instead of polling. It's only needed for long-lived agent services that listen on a webhook endpoint; interactive agents can ignore it entirely.

Net change: **+~3.4k / -~3k lines** across 46 files. Detailed commit log below.

## Motivation

OpenClaw introduced a separate process, custom install flow (`canonry agent setup`), and a pinned external dependency tree that didn't match canonry's "agent-first, CLI-native" direction. The native path:

- No separate process — Aero lives inside `canonry serve`.
- No install dance — `canonry agent ask <project> "…"` works as soon as an API key for any supported LLM is configured.
- First-class streaming, parallel tool execution, and an extensible tool registry courtesy of `pi-agent-core`.
- Hybrid persistence: live `Agent` instance in-memory per project, durable transcript + follow-up queue in a new `agent_sessions` table.

### CLI — built-in Aero

```bash
canonry agent ask <project> "<prompt>" [--provider anthropic|openai|google|zai] [--model <id>] [--format json]
```

Conversations persist per project — subsequent `agent ask` invocations continue the same rolling thread.

### CLI — external-agent push webhook (unchanged, just re-housed)

```bash
canonry agent attach <project> --url <webhook-url>
canonry agent detach <project>
```

`attach` registers a notification row (`source: 'agent'`) subscribed to `run.completed`, `insight.critical`, `insight.high`, `citation.gained`. Canonry POSTs each event to the URL. For human-driven agents like Claude Code in a terminal, you don't need this — just run the regular CLI/API. For long-lived agent services that want push instead of polling, this is the hook.

### Dashboard
Bottom command bar on every project-scoped route. Cursor/Raycast shape — not a dedicated chat panel. Starter buttons for common ops (Status, Top insights, Last failed run, Schedule). SSE-streamed via `POST /api/v1/projects/:name/agent/prompt`.

### Proactive Aero
`RunCoordinator` now enqueues a synthesized `[system]` follow-up into the session after every `run.completed`; `SessionRegistry.drainNow` wakes the agent unprompted. First end-to-end validation: triggered a failing sweep, Aero woke up, called `get_run` + `get_health`, diagnosed the Gemini API key issue, wrote the analysis back to the transcript — zero user intervention.

### Tools (13)

**Read (7):** `get_status`, `get_health`, `get_timeline`, `get_insights`, `list_keywords`, `list_competitors`, `get_run`.

**Write (6):** `run_sweep`, `dismiss_insight`, `add_keywords`, `add_competitors`, `update_schedule`, `attach_agent_webhook`. Additive-only — Aero recommends removals in prose, doesn't enact them. Tool intents surface via `tool_execution_start` events so users see exactly what fires.

### Persistence
New `agent_sessions` table (migration v38). One row per project (UNIQUE on `project_id`), stores transcript + follow-up queue + chosen model. Survives `canonry serve` restarts. Docs: `docs/data-model.md`.

## Commit log

```
573cb96  chore(agent): remove OpenClaw gateway and bundled runtime
e365938  feat(agent): pi-agent-core integration scaffold
7d04bf0  feat(agent): port 3 read tools (pattern-prover)
3be5b63  feat(agent): canonry agent ask CLI
9cfce4e  feat(agent): z.ai (GLM) provider + env-var API key fallback
ab54be2  feat(agent): batch-port remaining read tools
296938d  feat(agent): write tools (6)
e3e9532  feat(agent): persistent session registry + agent_sessions table
593c000  feat(agent): proactive Aero — registry-driven CLI + RunCoordinator wake-up
14769e1  fix(agent): dup-message + agent_end drain hook
cafb0b2  feat(agent): SSE agent routes
bdbd45b  feat(web): Aero bottom command bar
57b239a  docs(agent): AGENTS.md + README + canonry-cli skill sync
0da4f70  fix(agent): abort listener — response-side close, not request-side
39f3a67  fix(agent): code-review — auth, listener leaks, reader abort
```

## Breaking changes (2.0.0 → 2.6.2)

Removed CLI commands: `canonry agent setup`, `agent start`, `agent stop`, `agent status`, `agent reset`.

Removed config fields: `agent.{binary,profile,autoStart,gatewayPort}`. Only `agent.mode` remains (reserved).

`canonry agent attach` now requires `--url <webhook-url>` instead of deriving the URL from the former `agent.gatewayPort`.

Users with an orphaned `~/.openclaw-aero/` dir get a one-time boot-time warning and can delete it safely.

## Test plan

- [x] Workspace: typecheck, lint, 1000 tests green (all packages)
- [x] CLI dogfood: `canonry agent ask ainyc "…"` against a real project — Aero uses parallel tool execution, iterates on truncated tool results, produces actionable output
- [x] CLI threading: two `agent ask` invocations in sequence — Aero remembers context across them
- [x] Write tool live: `run_sweep` from Aero created a real queued run in canonry DB
- [x] Proactive wake-up: failed sweep → Aero analyzed unprompted → wrote diagnosis + fix recommendation to transcript
- [x] Duplicate-message fix: after `queueFollowUp`/`drainNow` race fix, exactly one `[system]` entry per run.completed
- [x] SSE flow: `POST /agent/prompt` streams `stream_open` → `agent_start` → per-token `message_update` → `tool_execution_*` → `message_end` → `stream_close`
- [x] Auth: unauthenticated `GET /agent/transcript` returns `401`; authenticated returns `200`
- [x] Abort listener fix: prior bug where `request.raw.on('close')` fired on POST body completion (not client disconnect) and canceled the agent immediately. Fixed by listening on `reply.raw.once('close')`.
- [ ] Live-in-browser manual QA of the AeroBar — build is clean, types are clean, tests pass, but I haven't opened the dashboard in Chrome on this branch. Suggested manual check: open `http://localhost:4100/projects/<name>`, click the bar, try a starter prompt, verify streaming text + tool pills render.

## Follow-ups for later PRs (not in this one)

- Confirmation gating for write tools in UI mode (`beforeToolCall` policy) — currently reads and writes fire without interstitial confirmation.
- GA4 / Bing read tools — current set is canonry-core only.
- `cancel_run` tool — useful when Aero fires a sweep the user wants to abort.
- Skill hot-reload — aero/SKILL.md is loaded once at session construction; changes require `evict()`.

## Before merging

Two cleanup items I'd happily do as a follow-up PR but want to flag:

- `packages/canonry/assets/assets/index-*.{js,css}` — generated build artifacts are committed (pre-existing; not introduced by this PR but the hashed filenames churn on every rebuild). Adding them to `.gitignore` and untracking is a ~3-line follow-up.
- One reviewer-flagged nit on `tools.ts` — the `as unknown as AgentTool` casts in `buildReadTools`/`buildWriteTools` could use a single type-safe helper. Cosmetic; left as-is to keep diff focused.


arberx:
**P1-1** (CLI bypass): `agent-ask.ts` now POSTs to `/api/v1/projects/:name/agent/prompt` with `scope: 'all'` and parses the SSE stream. Same session store as the dashboard — one execution path.

**P1-2** (parallel prompts): `state.isStreaming` guard returns `409 AGENT_BUSY` (new error code). Verified with simultaneous curls — first=200, second=409.

**P1-3** (dashboard write blast radius): split `toolScope`. Dashboard POST defaults to `read-only` (7 tools); CLI passes `scope: 'all'` (13 tools). Live verification: asked both surfaces to list available tools; dashboard listed 7, CLI listed 13. Tools swap per-request on the cached `Agent` — safe now that we 409 concurrent requests.

**P2-4** (provider/model ignored): explicit `preferences.provider` / `preferences.modelId` in `getOrCreate` now override the persisted values AND are written back, so subsequent invocations use the new model.

**P2-5** (proactive invisible in UI): `AeroBar` polls `fetchAeroTranscript` every 15s while open and no prompt is in flight. Server-initiated turns surface within one cycle.

**Open questions:**
- CLI parity: added `canonry agent transcript <project>` and `canonry agent reset <project>` (GET + DELETE).
- OpenAPI coverage: new `canonryLocalRouteCatalog` in `packages/api-routes/src/openapi.ts`, activated only when the caller passes `includeCanonryLocal: true` (canonry does; the shared contract test doesn't, so it still passes).

Full workspace still 1000 tests green. Bumped to 2.7.0 — new error code + CLI commands + dashboard scope default are user-visible changes.

arberx:
Pass 3 in `59f7de5`:

**P1-1** (tool swap before busy check): introduced `SessionRegistry.acquireForTurn` — busy guard fires first, then scope/model align only on the idle path. `getOrCreate` is now pure (never mutates a cached Agent). Agent-routes + drainNow both call `acquireForTurn`; drainNow catches AGENT_BUSY and leaves pending for the `agent_end` hook. Regression test asserts `state.tools` is unchanged when the guard throws.

**P1-2** (CLI bypass of ApiClient + CliError): `agent-ask.ts` and `agent-transcript.ts` now go through `createApiClient()`. Added `ApiClient.streamPost(path, body, signal)` that preserves the `/health` basePath probe + bearer auth + structured-error path and returns a `Response` for SSE consumption. Added `getAgentTranscript()` / `resetAgentTranscript()` for the transcript subcommands. CLI errors flow through `printCliError` with proper `CliError.exitCode`, matching the repo's 0/1/2 contract. The probe re-engages, so reverse-proxied deployments without a local basePath config now resolve correctly.

**P2-1** (hot-session model overrides): `acquireForTurn` aligns `state.model` on cached Agents when `preferences.provider` or `preferences.modelId` differs, and persists to the DB row. Regression test confirms a cached `anthropic` agent switches to `zai/glm-5.1` when the next request passes `provider: 'zai'`.

**P2-2** (2.0.0 vs 2.7.0): intentional rollback. The repo owner asked me to hold the version through the whole branch — upstream main is still on 1.x; this feature ships as a single 2.0.0 release. Earlier PR comment about 2.7.0 is stale; the final target is 2.0.0.

Live-verified:
- CLI `agent ask` routes through `ApiClient.streamPost` — Aero replied on the first try.
- Parallel curls: first=200, second=409 with `AGENT_BUSY`.

1015 tests green (+3 for `acquireForTurn` regression coverage).

packages/canonry/src/agent/session-registry.ts
```diff
@@ -0,0 +1,424 @@
...
+   */
+  async drainNow(projectName: string): Promise<void> {
+    if ((this.pending.get(projectName) ?? []).length === 0) return
+    try {
+      let agent: Agent
+      try {
+        // acquireForTurn does the busy check in the registry — if the agent
+        // is mid-stream we leave pending alone and let `agent_end` drain
+        // hook pick it up. Pi's AppError surfaces as `AGENT_BUSY`.
+        agent = this.acquireForTurn(projectName)
```
arberx: **Tool-scope escalation during proactive drain** — `drainNow` calls `acquireForTurn(projectName)` with no preferences, so `alignScope` receives `preferences?.toolScope ?? 'all'` and unconditionally rebuilds `agent.state.tools` with the full write set:

```ts
// acquireForTurn (~line 186)
this.alignScope(projectName, agent, preferences?.toolScope ?? 'all')

// alignScope (~line 193) — fires when current scope is 'read-only' and wantScope is 'all'
if (this.scopes.get(projectName) === wantScope) return
agent.state.tools = wantScope === 'read-only' ? buildReadTools(...) : buildAllTools(...)
this.scopes.set(projectName, wantScope)
```

Repro: dashboard opens a session in read-only (AeroBar default), a `run.completed` fires, `RunCoordinator` → `drainNow` → the synthesized `[system]` follow-up now runs with write tools available (`run_sweep`, `dismiss_insight`, `add_keywords`, `add_competitors`, `update_schedule`, `attach_agent_webhook`). The UI's read-only gate is silently upgraded for proactive turns without user intent.

Fix: either persist/track the session's "requested" scope separately from the live-Agent scope and pass `{ toolScope: requestedScope }` when draining, or hard-pin proactive drains to `'read-only'` since they're system-triggered.

Rule: [AGENTS.md — agent-first principles](https://github.com/AINYC/canonry/blob/main/AGENTS.md) — write tools must stay behind explicit user intent; the PR description itself states "Tool intents surface via `tool_execution_start` events so users see exactly what fires", which the escalation path undermines.

docs/data-model.md
```diff
@@ -90,6 +90,12 @@ erDiagram
 | **api_keys** | API authentication. Unique: `keyHash` |
 | **usage_counters** | Rate limiting and usage tracking. Unique: `(scope, period, metric)` |
 
+### Agent
+
+| Table | Purpose |
```
arberx, line 97: **ER diagram not updated.** The prose table-group was added here, but the Mermaid `erDiagram` block at the top of this file (lines ~8–34) still has no entry for `agent_sessions`. The root [AGENTS.md "Keeping Documentation Current" table](https://github.com/AINYC/canonry/blob/main/AGENTS.md#keeping-documentation-current) requires updating *both*:

> | Add a new table or column in `packages/db/src/schema.ts` | Update `docs/data-model.md` (**ER diagram** + **table groups**) |

Expected addition to the `erDiagram` block — the schema has `UNIQUE(projectId)`, so the 1:1 form matches the `schedules` entry:

```
projects ||--o| agent_sessions : "has (1:1)"
```

skills/canonry-setup/references/canonry-cli.md
```diff
@@ -298,51 +298,44 @@ canonry sitemap inspect <project>
...
-canonry agent detach <project> --format json     # JSON output
+# One-shot turn — Aero picks its own tools, streams events to stdout.
+canonry agent ask <project> "<prompt>"
+canonry agent ask <project> "<prompt>" --format json      # JSON event stream
+
+# Select a specific provider / model (otherwise auto-detected from config).
+canonry agent ask <project> "<prompt>" --provider anthropic --model claude-opus-4-7
+canonry agent ask <project> "<prompt>" --provider zai      --model glm-5.1
+canonry agent ask <project> "<prompt>" --provider openai
+canonry agent ask <project> "<prompt>" --provider google
```
arberx, line 316: **`--scope` flag missing from the skill reference.** `canonry agent ask` in this PR adds a `--scope all|read-only` flag (registered in `packages/canonry/src/cli-commands/agent.ts` ~line 14–44), but the skill reference only documents `--provider`, `--model`, and `--format json`. The flag changes tool-surface behavior (read-only vs. full write), so agents generated via this skill will silently default to one side without knowing the lever exists.

Per [root CLAUDE.md → "Skills Maintenance"](https://github.com/AINYC/canonry/blob/main/CLAUDE.md#skills-maintenance):

> - **New CLI command** → add it to `skills/canonry-setup/references/canonry-cli.md`

A new flag on an existing command lives under the same rule. Add a line such as:

```bash
canonry agent ask <project> "<prompt>" --scope read-only   # restrict to read tools (default for dashboard)
canonry agent ask <project> "<prompt>" --scope all         # expose write tools (default for CLI)
```

PR #396, opened 2026-05-02
fix(tests): declare jsdom environment in vitest config instead of inline directives

arberx:
## Problem

vitest worker pool couldn't resolve `jsdom` for `apps/web` test files that use `// @vitest-environment jsdom` inline directives. PNPM's strict dependency resolution prevented the root-level vitest worker from finding jsdom in `apps/web/node_modules`.

## Solution

Move the environment declaration from per-file directives to `apps/web/vitest.config.ts`:

- Set `environment: 'jsdom'` in `apps/web/vitest.config.ts`
- Remove `// @vitest-environment jsdom` directives from 3 `.tsx` test files
- Revert root `jsdom` devDependency (not needed — jsdom stays in `apps/web` where it belongs)

## Verification

```
Test Files  15 passed (15)
     Tests  169 passed (169)
```

No production dependency changes.

### DIFF
--- a/apps/web/test/router.test.tsx
+++ b/apps/web/test/router.test.tsx
@@ -1,4 +1,3 @@
-// @vitest-environment jsdom
 import { test, expect } from 'vitest'
 import React from 'react'
 import { render, waitFor, act } from '@testing-library/react'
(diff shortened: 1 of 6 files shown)


arberx:
Superseded — environment: 'jsdom' in vitest config has the same PNPM resolution problem. Vitest resolves environment packages from workspace root, not project directory. Reverting to inline directives + root jsdom devDep (#395).

PR #399, opened 2026-05-02
fix(matching): gate normalized brand candidate by MIN_BRAND_KEY_LENGTH

evolv3ai:
## Summary

`brandNormalizedCandidates(displayName)` returned the normalized form of a short displayName (e.g. \`\"LI\"\` → \`\"li\"\`, \`\"Inc\"\` → \`\"inc\"\`) without any length guard, so a 2–3 character candidate substring-matched inside common English vocabulary on the \`matchesNormalized\` path:

- \`\"li\"\` inside \`tile\`, \`vinyl\`, \`polished\`, \`installation\`, \`compliance\`
- \`\"inc\"\` inside \`incident\`, \`include\`, \`since\`

The existing \`MIN_BRAND_KEY_LENGTH = 6\` guard already protected the **brand-key** path (line 80) and the **stripped** variant of the normalized path (line 194), but the **original** normalized candidate dropped through unguarded.

This PR adds a symmetric guard on the original candidate. Short brands still detect correctly through the protected paths: domain mention, distinctive tokens (word-boundary regex), and the brand-key path.

The change is a cousin of #392 — same family of false-positive substring matches in answer-visibility.

## Field evidence

A project with \`displayName = \"LI\"\` (a 2-letter abbreviation) had \`answerMentioned = true\` on **89 of 90** query snapshots across three sweeps — every commercial-flooring answer the providers returned contained at least one of \`tile\`, \`vinyl\`, \`polished\`, or \`installation\`, and each of those words contains the substring \"li\".

After this fix (verified locally via \`canonry backfill answer-mentions\`):

| state | mentioned / 90 |
|---|---:|
| before | 89 |
| after  | 5 |

The post-fix count matches an independent from-scratch regex audit on the raw \`answerText\` for \"Larry's Interiors\" / \"larrysinteriors.com\".

## Tests

- New: \`extractAnswerMentions does not substring-match short normalized display names inside common words\` — covers the LI / tile case plus sanity checks that domain match and full brand match still work.
- Tightened: the existing \`does not strip a classifier when only the classifier itself remains\` test now asserts \`mentioned: false\` (it previously documented this as a \"pre-existing limitation of the strict path for very short brand names\" — that limitation is now closed).

\`pnpm run test\` — 1728 passed (165 files) ✅
\`pnpm run typecheck\` ✅
\`pnpm run lint\` ✅

## Test plan

- [x] Add regression test for short displayName substring false-positives
- [x] Tighten existing \"Inc / incident\" test to assert the bug is closed
- [x] Verify \`pnpm run test\` passes across the full workspace
- [x] Verify \`pnpm run typecheck\` and \`pnpm run lint\` pass
- [x] Bump root + \`@ainyc/canonry\` version (3.2.7 → 3.2.8)
- [x] Field-verify with \`canonry backfill answer-mentions\` on the real project that surfaced this (89/90 → 5/90)


arberx:
Thanks for making a PR!

Good find, current logic is incorrect however I dont think this will solve it either....seems we need a word boundary regex.

PR #449, opened 2026-05-09
feat(scheduler): wire traffic-sync into the schedule system

arberx:
## Summary

Project schedules can now dispatch either the existing **answer-visibility** kind OR a new **traffic-sync** kind, keyed by `(project_id, kind)`. Closes the last item in the assessment-driven traffic-integration roadmap: scheduled traffic syncs, safe under the cross-sync dedupe ring buffer landed in #444.

## What you can do after this lands

```bash
# Visibility schedule (default kind, unchanged)
canonry schedule set my-project --cron "0 0 * * *"

# Traffic-sync schedule
canonry schedule set my-project --kind traffic-sync --source <source-id> --cron "0 */6 * * *"

# Both can coexist on the same project
canonry schedule show my-project --kind answer-visibility
canonry schedule show my-project --kind traffic-sync
```

## Schema (migration v53)

`schedules` gains:
- `kind TEXT NOT NULL DEFAULT 'answer-visibility'` — which RunKind to dispatch
- `source_id TEXT` — required when kind='traffic-sync', null otherwise

The legacy `UNIQUE(project_id)` was an **inline table constraint** (not a droppable index), so the migration uses SQLite's canonical table-rebuild pattern inside the runner's single transaction:

```sql
CREATE TABLE schedules_v53 (... UNIQUE(project_id, kind));
INSERT INTO schedules_v53 (...) SELECT ..., 'answer-visibility', NULL FROM schedules;
DROP TABLE schedules;
ALTER TABLE schedules_v53 RENAME TO schedules;
```

Existing rows backfill to `kind='answer-visibility'`. New install bootstrap (MIGRATION_SQL) is unchanged — fresh DBs run the v53 migration which converts the legacy schema in one pass.

## Contracts

- `SchedulableRunKind = 'answer-visibility' | 'traffic-sync'` + `SchedulableRunKinds` enum-constants object (per the "Enum Constants (Critical)" rule).
- `ScheduleDto` gains `kind` and `sourceId`.
- `scheduleUpsertRequestSchema` accepts optional `kind` and `sourceId`. Callers that omit `kind` default to `'answer-visibility'` — no breaking change.

## Routes

PUT/GET/DELETE on `/projects/:name/schedule` accept optional `?kind=` query (or body `kind` field on PUT). Per-kind invariants enforced at the route layer:

- `traffic-sync` requires `sourceId`, validates the source exists for the project, rejects `providers` (only meaningful for visibility runs)
- Non-traffic-sync rejects `sourceId`

`onScheduleUpdated` callback signature expands to `(action, projectId, kind)`. `apply.ts` (the config-as-code path) passes `'answer-visibility'` explicitly — YAML-driven applies don't manage traffic schedules.

## Scheduler

Tasks keyed by `(projectId, kind)`. Trigger logic branches:

- **answer-visibility**: existing flow → `queueRunIfProjectIdle` → `onRunCreated` → JobRunner
- **traffic-sync**: validates sourceId, calls a new `onTrafficSyncRequested(projectName, sourceId)` callback that the canonry server wires to `apiClient.trafficSync(...)` — reusing the exact same in-process flow that runs the cross-sync dedupe, writes the run row, and emits `traffic.synced` telemetry (per #448)

`removeAllForProject(projectId)` clears all kinds for a project (used on project delete).

## CLI

`canonry schedule set/show/enable/disable/remove` accept `--kind` and `--source`. Default kind is `'answer-visibility'` so existing scripts keep working unchanged.

## Test plan

- [x] 3 new scheduler tests (both-kinds-coexist, traffic-sync trigger fires the right callback, traffic-sync silently skips when sourceId missing)
- [x] 1 existing test updated for new triggerRun signature
- [x] All 2,213 workspace tests pass
- [x] `pnpm typecheck` clean across workspace
- [x] `pnpm lint` clean across workspace
- [x] Migration v53 verified: fresh-DB scheduler tests register answer-visibility AND traffic-sync schedules without UNIQUE constraint violation

## Versioning

`4.16.2 → 4.16.3` (patch — additive scheduling surface, no breaking changes to existing API).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -59,6 +59,14 @@ canonry report <project>                         # client-facing AEO report → 
 canonry report <project> --output dist/aeo.html
 canonry report <project> --format json           # raw report payload to stdout
 
+# Schedules — one row per (project, kind) where kind ∈ {answer-visibility, traffic-sync}
+canonry schedule set <project> --preset daily                                                # answer-visibility (default kind)
+canonry schedule set <project> --kind traffic-sync --cron "*/15 * * * *" --source <id>       # traffic-sync (sourceId required)
+canonry schedule show <project> [--kind answer-visibility|traffic-sync]                      # default kind is answer-visibility
+canonry schedule enable  <project> [--kind ...]
+canonry schedule disable <project> [--kind ...]
+canonry schedule remove  <project> [--kind ...]                                              # delete the schedule for that kind
+
 # Agent layer
 canonry agent ask <project> "<prompt>"               # one-shot turn against built-in Aero
 canonry agent ask <project> "<prompt>" --provider zai --format json
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -12,7 +12,7 @@ The web dashboard follows a dark, professional analytics aesthetic inspired by *
 
 ### Color & Theme
 - Background: `bg-zinc-950`. Cards/surfaces: `bg-zinc-900/30` with `border-zinc-800/60`.
-- Font: **Manrope** (400–800 weights), `text-zinc-50` primary, `text-zinc-400` secondary, `text-zinc-500`/`text-zinc-600` for labels.
+- Font: **Geist Sans** (400–800 weights) for UI text, **Geist Mono** for code/numerics. Globally enabled OpenType features `cv11`, `ss01`, `ss03` for sharper i/l/I/0 disambiguation. Headings tighten tracking (`-0.015em`, `-0.02em` on h1). `text-zinc-50` primary, `text-zinc-400` secondary, `text-zinc-500`/`text-zinc-600` for labels.
 - Tone colors: **positive** = emerald, **caution** = amber, **negative** = rose, **neutral** = zinc.
 - No decorative background gradients. Keep it clean and flat.
 
--- a/apps/web/src/pages/TrafficSourceDetailPage.tsx
+++ b/apps/web/src/pages/TrafficSourceDetailPage.tsx
@@ -126,13 +126,13 @@ export function TrafficSourceDetailPage() {
   }
 
   return (
-    <div className="page-container">
+    <div className="page-container space-y-8">
       <div className="page-header">
         <div className="page-header-left">
           <Link to="/traffic" className="inline-flex items-center gap-1 text-xs text-zinc-500 hover:text-zinc-200">
             <ArrowLeft className="size-3" /> All sources
           </Link>
-          <h1 className="page-title mt-1">{detail.displayName}</h1>
+          <h1 className="page-title mt-2">{detail.displayName}</h1>
           <p className="page-subtitle">
             {detail.sourceType} · project <span className="text-zinc-300">{projectName}</span> ·
             <span className="ml-1 font-mono text-zinc-400">{detail.id}</span>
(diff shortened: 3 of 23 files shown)


arberx:
Review addressed in `f4e3dcac`. Full validation: typecheck + lint + 2220 tests pass.

## 🔴 Critical
**1. apply.ts kind-scoping** — fixed. All three schedules SELECT/UPDATE/DELETE now scope on `kind = SchedulableRunKinds['answer-visibility']` and the INSERT explicitly sets that kind. New tests in `packages/api-routes/test/schedules.test.ts` seed both kinds and assert the traffic-sync row is preserved through both branches (no spec.schedule, and spec.schedule set).

## 🟠 High
**2. OpenAPI drift** — added `kind` query parameter to PUT/GET/DELETE `/projects/{name}/schedule`, and `kind` + `sourceId` to the PUT body schema, plus a 400 response on PUT for invariant violations.

**3. MCP parity** — `canonry_schedule_get` and `canonry_schedule_delete` now use a new `scheduleReadInputSchema` that accepts `{ project, kind? }` and forwards `kind` to `client.getSchedule`/`deleteSchedule`. `mcp-registry.test.ts` now asserts both call shapes (with and without kind).

## 🟡 Medium
**4. UI parity gap** — **deferring to a follow-up PR**. The SPA (`apps/web/src/api.ts:512/526/533` + `ScheduleSection.tsx`) currently only manages the answer-visibility schedule. Adding a kind tab/picker is a non-trivial UI change with its own design surface and tests; bundling it into this PR would expand scope and slow the critical fix landing. Will file a follow-up issue with the specific files to touch (`api.ts`, `ScheduleSection.tsx`, plus any selectors that need a kind filter).

**5. Test coverage gaps** — covered. New `schedules.test.ts` adds:
- `sourceId` required when kind=traffic-sync
- `sourceId` rejected for answer-visibility
- `providers` rejected for traffic-sync
- cross-project `sourceId` rejected (404)
- happy-path round-trip (PUT → GET `?kind=traffic-sync` → DELETE)

## 🟢 Nits
**6. Redundant UNIQUE in v53** — fixed. Inline `UNIQUE(project_id, kind)` removed from `CREATE TABLE schedules_v53`; the explicit `CREATE UNIQUE INDEX idx_schedules_project_kind` remains as the canonical (drizzle-side-named) index.

**7. AGENTS.md commands block** — added a new `# Schedules` block with `--kind` / `--source` examples for `set`, `show`, `enable`, `disable`, `remove`.

**8. Version bump** — bumped to **4.17.0** (was 4.16.3 ← actually 4.15.2, branch was behind main). Minor reflects the new schedulable run kind, new contract field, new CLI flags, MCP behavior change, and migration v53.

## Commit
`f4e3dcac fix(scheduler): scope apply.ts schedule writes by kind + MCP/OpenAPI parity`

Files changed: 9 files, +376/-25. Net additions are mostly the new tests file.

arberx:
**Live UI test caught a second bug.** Building 4.17.0 locally and exercising the new traffic-sync schedule UI surfaced a migration boot-loop:

`MIGRATION_SQL` runs on every boot before versioned migrations, and it was still creating the legacy `idx_schedules_project` (UNIQUE single-column on `project_id`). v53 dropped that index in Phase 2, but on the very next boot Phase 1 re-created it. After that, `canonry schedule set demo --kind traffic-sync ...` failed with `UNIQUE constraint failed: schedules.project_id` — exactly when the new feature is supposed to work.

Repro from the local exercise:
1. Fresh `canonry init` + `canonry serve` (4.17.0)
2. Saved an answer-visibility schedule via the dashboard ✅
3. Restarted the server (any reason — reboot, redeploy)
4. `canonry schedule set demo --kind traffic-sync --cron "*/15 * * * *" --source <id>` → 500, `UNIQUE constraint failed: schedules.project_id`

Fix in `d9a3dd4c`:
- Removed `CREATE UNIQUE INDEX idx_schedules_project` from `MIGRATION_SQL` so it stops resurrecting the legacy index every boot.
- Added migration **v54 `drop-resurrected-schedules-project-index`** which `DROP INDEX IF EXISTS idx_schedules_project` to clean up DBs that already booted past v53 with the resurrected index.
- New regression test `packages/db/test/schedules-migration.test.ts` calls `migrate()` twice in a row (simulating a reboot) and asserts both schedule kinds remain insertable. Tests fail with the same SqliteError without the fix.

Version bumped 4.17.0 → 4.17.1 (patch — boot-time migration bug on top of the v53 minor).

Also confirmed live in the same exercise:
- The dashboard's Server-Activity / `/traffic` page renders correctly with hourly chart, kind filters, and event-row table.
- The Cloud Run connect modal is well-formed and the security note about `~/.canonry/config.yaml` is shown.
- After this fix, `canonry apply` with no `spec.schedule` correctly **preserves** the traffic-sync schedule (the original critical fix verified end-to-end).
- The deferred UI parity gap (#4) is also confirmed live: the project page's "Scheduled runs" section says "No schedule configured" even when a `traffic-sync` row exists in the DB, because the SPA fetches `/schedule` without `?kind=`. Will file the follow-up issue.

PR #479, opened 2026-05-13
fix(web): render every location after multi-location fan-out (#477)

### LINKED ISSUE #477: bug(multi-location): citation tracking renders only one location after --all-locations fan-out; latest-run resolution non-deterministic
## Summary

For projects with multiple configured `locations`, a sweep run with `--all-locations` fans out into **one run per location, all sharing the same `createdAt` timestamp**. The project page's client-side "latest run" resolution in `apps/web/src/queries/use-dashboard.ts` then picks one of the runs non-deterministically and uses only its snapshots — the other location's data is silently dropped from every downstream surface.

Net effect for a 2-location project (`azcoatings`: `florida` + `michigan`, both runs at `2026-05-13T17:23:20.060Z`):

- Only **one** location chip renders in the citation tracking filter row (florida wins consistently in our data, but the choice is dictated by the order `fetchAllRuns` returns runs in, combined with stable sort).
- Selecting the visible chip filters snapshots but **not queries** — geo-named queries (`polyurea roof coating michigan`) still appear as rows under `florida` filter because the florida run did sweep them too (queries are project-scoped today). This is a real read-path bug *and* a UX confusion that the enhancement in #478 would resolve at the data-model layer.
- The other location's snapshots are invisible in the project page — no chip, no rows, no answer text.

Reproduced on canonry `4.26.0` against `azcoatings` (revision was published with the run-trigger `queries[]` scope from #476 just merged).

## Reproduction

1. Create a project with ≥2 configured locations.
2. Fire `canonry run <project> --all-locations --wait`. Two runs are created with the **same `createdAt`** (the API handler inserts both inside a single `now = new Date().toISOString()` block at `packages/api-routes/src/runs.ts:86-100`).
3. Open the project page. Scroll to `Visibility evidence → Query citation tracking`.

## Observed behavior

- **(1) Only one location chip renders.** Chip row gates each chip on `locationLabelsInEvidence.has(loc.label)` in `apps/web/src/pages/ProjectPage.tsx:1607-1619`. Since the evidence set is built from a single `latestRunDetail`, the location-label set only contains one label.
- **(2) Tiebreak is client-side and order-dependent.** Originally I assumed this was in the server's `/runs/latest` endpoint — that's wrong. The SPA never calls `/runs/latest` for this path. It calls `fetchAllRuns`, then ranks client-side:
  ```ts
  // apps/web/src/queries/use-dashboard.ts:57-60
  const completedRuns = projectRuns
    .filter(r => (r.status === 'completed' || r.status === 'partial') && r.kind === 'answer-visibility')
    .sort((a, b) => b.createdAt.localeCompare(a.createdAt))
  const latestRunId = completedRuns[0]?.id
  ```
  `localeCompare` returns 0 for identical ISO strings, and `Array.prototype.sort` is stable in modern V8, so order is preserved from `fetchAllRuns`. Florida winning the tie is just a function of what order the server emits runs in `GET /runs` — not anything semantic.
- **(3) The "Compare" chip cannot appear with multi-location fan-out** because the precondition `distinctLocationsWithEvidence.length > 1` is never satisfied — evidence only contains one location's data.
- **(4) Location filter filters snapshots, not queries.** `filteredEvidence` filters rows on `e.location === locationFilter` (`ProjectPage.tsx:1152`). The query `polyurea roof coating michigan` shows up under `florida` filter because the florida run *did* produce a florida-grounded snapshot for that query (queries are project-scoped, not query-scoped — see #478 for the proposed fix). The florida snapshot's answer mentions Michigan because the query *text* asks about Michigan. This is correct LLM behavior, exposed as a UX bug only because of the missing per-query geo scope.
- **(5) `GET /projects/:name/timeline?location=<label>` returns entries whose run rows carry `location: null`.** A timeline query for `?location=florida` on azcoatings returns 6 query entries; the `latestRunLocation` on the first entry is `null` despite the filter. Either the filter isn't being applied server-side or the response strips the field from the projection. Needs separate verification.

## Expected behavior

- All configured locations should be selectable in the filter row, regardless of which single run won the latest-run tiebreaker.
- Switching the chip should swap the evidence rows to that location's snapshots — answer text, citation states, cited domains, all per-location.
- The "Compare" chip should appear whenever ≥2 configured locations have any data.
- `GET /timeline?location=X` should return runs with `location === X`, never `null` or a different label.
- (Out of scope here, but exposed in (4)) — geo-named queries should be opt-out-able from non-matching locations via #478.

## Root cause analysis

The data model is correct. `runs` and `query_snapshots` both carry `location`, and per-location snapshots persist accurately. The breakage is at the **client-side read-aggregation layer**, with one related server-side issue and one related cross-cutting issue:

### A. Client-side latest-run ranking is single-row (`use-dashboard.ts`)

The SPA fetches all runs, ranks by `createdAt`, and takes `[0]`. With same-timestamp fan-out, this collapses to one location.

```ts
// apps/web/src/queries/use-dashboard.ts:55-72
const projectRuns = allRuns.filter(r => r.projectId === project.id)
const completedRuns = projectRuns
  .filter(r => (r.status === 'completed' || r.status === 'partial') && r.kind === 'answer-visibility')
  .sort((a, b) => b.createdAt.localeCompare(a.createdAt))

const latestRunId = completedRuns[0]?.id
// ...
const [..., latestRunDetail, ...] = await Promise.all([
  ...,
  latestRunId ? fetchRunDetail(latestRunId).catch(() => null) : Promise.resolve(null),
  ...,
])
```

### B. Evidence builder assumes a single latest run (`build-dashboard.ts`)

`buildEvidenceFromTimeline` reads `latestRunDetail.snapshots`, keys by `query::provider`, and emits one row per (query × provider). With multi-location fan-out the second location's snapshots are dropped — both their location chip and their cell-level data.

```ts
// apps/web/src/build-dashboard.ts:155-274
if (latestRunDetail) {
  const snapshotsByKey = new Map(...)
  for (const snap of latestRunDetail.snapshots) {
    if (snap.query) snapshotsByKey.set(`${snap.query}::${snap.provider}`, snap)
  }
  // ...
  for (const provider of providers) {
    const snap = snapshotsByKey.get(`${entry.query}::${provider}`)
    results.push({ ..., location: snap?.location ?? null, ... })
  }
}
```

### C. Filter chip gating is data-driven, not config-driven (`ProjectPage.tsx`)

Configured locations with no evidence rows are chip-less. Combined with (A) and (B), every fan-out leaves N-1 locations un-selectable.

```tsx
// apps/web/src/pages/ProjectPage.tsx:1607-1619
{model.project.locations.map((loc: { label: string }) => (
  locationLabelsInEvidence.has(loc.label) && (
    <button>{loc.label}</button>
  )
))}
```

### D. Timeline endpoint location filter (server-side)

`GET /projects/:name/timeline?location=florida` returns rows whose runs carry `location: null`. Either the WHERE clause is missing/wrong or the projection strips `location`. Needs server-side trace; a regression test would lock it down.

### E. `/runs/latest` tiebreak is also non-deterministic (separate hardening)

This is **not** the cause of the SPA bug (the SPA doesn't call it), but it's a real latent issue. Other consumers — MCP `canonry_runs_latest`, any external agent webhook subscribers, future CLI helpers — call `/runs/latest` and will silently get a different run after a backup/restore or DB rebuild. One-line fix: add `id DESC` as a secondary tiebreak.

## Suggested fixes

1. **`apps/web/src/queries/use-dashboard.ts`** — detect same-timestamp completed answer-visibility runs (group `completedRuns` by `createdAt` ms). For the top group, fetch run detail for **every** run via parallel `fetchRunDetail` calls; pass an array (`latestRunDetails[]`) down to `buildDashboard`.

2. **`apps/web/src/build-dashboard.ts`** — `buildEvidenceFromTimeline` accepts `latestRunDetails: ApiRunDetail[]` instead of a single detail. Two shape options:
   - **(2a)** Emit one evidence row per (query × provider × location). Cleanest data model but expands the table.
   - **(2b)** Emit one evidence row per (query × provider) carrying a `snapshotsByLocation: Map<string, Snapshot>` field. The displayed "current state" (answerSnippet, citationState, citedDomains) is swapped at render time when the user clicks a chip. Less ergonomic at the data layer but minimal changes to the table UI.
   Recommend **(2b)** for less churn in `EvidenceTable` and `CitationInsightVm`.

3. **`apps/web/src/pages/ProjectPage.tsx`** — drop the `locationLabelsInEvidence.has(loc.label)` gate. Render chips from `model.project.locations` directly. If a location has no evidence rows, the filtered table renders an empty-state caption rather than the chip being hidden.

4. **`packages/api-routes/src/intelligence.ts`** (or wherever `/timeline` lives) — write a test asserting `GET /projects/:name/timeline?location=X` returns runs with `location === X`. Fix the WHERE or projection as needed.

5. **`packages/api-routes/src/runs.ts`** — add a secondary tiebreak to `/runs/latest`: `ORDER BY createdAt DESC, id DESC LIMIT 1`. Independent hardening, separate small PR.

6. **(deferred to #478)** Per-query location scope removes the snapshot-filter-vs-query-set confusion at the data layer. Doesn't replace this fix — even with per-query scope, multi-location runs still need correct read-path aggregation.

## Affected files

- `apps/web/src/queries/use-dashboard.ts` — fix #1
- `apps/web/src/build-dashboard.ts` — fix #2
- `apps/web/src/pages/ProjectPage.tsx` — fix #3
- `apps/web/src/view-models.ts` — likely widened for fix #2 (`CitationInsightVm` gains `snapshotsByLocation` if going with 2b)
- `packages/api-routes/src/intelligence.ts` (or wherever timeline endpoint lives) — fix #4
- `packages/api-routes/src/runs.ts` — fix #5

## Test plan once fixed

- 2-location project, fire `--all-locations` sweep. Both location chips render.
- Click each chip — evidence rows update with the corresponding run's snapshots, including answer text, citation states, cited domains.
- "Compare" chip appears and renders side-by-side across both locations for each (query × provider).
- `GET /projects/:name/timeline?location=florida` → all `latestRunLocation` fields == `florida`, never `null` or `michigan`.
- `GET /runs/latest` returns the same run ID across repeated calls and across DB restore from backup.

## Related

- Enhancement #478 — per-query location scope. Solves the snapshot-filter-vs-query-set confusion (observation #4) at the data layer by letting users tag geo-baked queries (like `polyurea roof coating michigan`) so they only sweep at the matching location. Complementary, not a replacement.

## Notes

Surfaced while validating the daily orchestrator's reactive-trigger logic — the orchestrator needs to know which location an AI crawler hit came from and trigger a scoped sweep for that specific location. The current read-path collapse made the UI confusing to investigate, but the underlying snapshot data is fine, so this is purely a SPA read-path fix (plus one small server-side hardening + one timeline-endpoint test).

🤖 Originally filed with incorrect fix locations (server-side `/runs/latest`). Updated after tracing the real path in `use-dashboard.ts`.

arberx:
## Summary

Fixes #477 — multi-location citation tracking now renders every configured location, not just whichever one the client-side latest-run resolver happened to pick first.

## Changes
- **`use-dashboard.ts` / `build-dashboard.ts`** — group same-timestamp completed answer-visibility runs, fetch run detail for the whole group, and emit evidence rows per (query × provider × location) so both Florida and Michigan survive the aggregate.
- **`ProjectPage.tsx`** — chips render from `model.project.locations` directly so a chip-less location is no longer hidden; `Compare` falls back to configured locations when evidence-only data fails the ≥2 threshold.
- **`api-routes/runs.ts`** — `/runs/latest` now orders by `createdAt DESC, id DESC` for a deterministic tiebreak across backup/restore.
- **`api-routes/history.ts`** — timeline run entries now expose `location` so callers can verify the `?location=X` filter the server already applied.
- **`canonry/commands/schedule.ts`** — drops the duplicate "Schedule:"/"Cron:" rows in `canonry schedule show` when only `--cron` is configured (separate user-reported CLI bug); "Schedule:" → "Preset:", only rendered when a preset is set.

## Test plan
- [x] New `build-dashboard` regression test covers the two-location fan-out scenario from the issue.
- [x] New `/runs/latest` tiebreak test asserts deterministic ordering across repeated calls with same `createdAt`.
- [x] New `/timeline?location=X` test asserts run entries carry the requested location.
- [x] New `printSchedule` unit tests cover cron-only, preset, and traffic-sync schedules.
- [x] Full workspace suite: 228 files / 2512 tests pass, `pnpm typecheck` + `pnpm lint` clean.

### DIFF
--- a/apps/web/src/api.ts
+++ b/apps/web/src/api.ts
@@ -224,35 +224,22 @@ export interface ApiCompetitor {
   createdAt: string
 }
 
+export interface ApiTimelineRunEntry {
+  runId: string
+  createdAt: string
+  citationState: string
+  transition: string
+  answerMentioned?: boolean
+  visibilityState?: string
+  visibilityTransition?: string
+  location?: string | null
+}
+
 export interface ApiTimelineEntry {
   query: string
-  runs: {
-    runId: string
-    createdAt: string
-    citationState: string
-    transition: string
-    answerMentioned?: boolean
-    visibilityState?: string
-    visibilityTransition?: string
-  }[]
-  providerRuns?: Record<string, {
-    runId: string
-    createdAt: string
-    citationState: string
-    transition: string
-    answerMentioned?: boolean
-    visibilityState?: string
-    visibilityTransition?: string
-  }[]>
-  modelRuns?: Record<string, {
-    runId: string
-    createdAt: string
-    citationState: string
-    transition: string
-    answerMentioned?: boolean
-    visibilityState?: string
-    visibilityTransition?: string
-  }[]>
+  runs: ApiTimelineRunEntry[]
+  providerRuns?: Record<string, ApiTimelineRunEntry[]>
+  modelRuns?: Record<string, ApiTimelineRunEntry[]>
 }
 
 export interface ApiAuditEntry {
--- a/apps/web/src/build-dashboard.ts
+++ b/apps/web/src/build-dashboard.ts
@@ -145,26 +145,35 @@ function summaryFromRun(run: ApiRun): string {
 function buildEvidenceFromTimeline(
   projectName: string,
   timeline: ApiTimelineEntry[],
-  latestRunDetail: ApiRunDetail | null,
+  latestRunDetails: ApiRunDetail[],
   savedQueries: ApiQuery[],
 ): CitationInsightVm[] {
   const results: CitationInsightVm[] = []
   let idx = 0
   const seenQueries = new Set<string>()
 
-  if (latestRunDetail) {
-    // Group snapshots by query+provider for multi-provider support
+  if (latestRunDetails.length > 0) {
+    // Multi-location runs fan out as one ApiRunDetail per location, all with the
+    // same `createdAt`. Bucket snapshots by (query × provider × location) so each
+    // location's evidence row survives the aggregate instead of being clobbered.
+    const allSnapshots = latestRunDetails.flatMap(r => r.snapshots)
     const snapshotsByKey = new Map<string, ApiRunDetail['snapshots'][number]>()
-    for (const snap of latestRunDetail.snapshots) {
+    for (const snap of allSnapshots) {
       if (snap.query) {
-        const key = `${snap.query}::${snap.provider}`
+        const key = `${snap.query}::${snap.provider}::${snap.location ?? ''}`
         snapshotsByKey.set(key, snap)
       }
     }
 
+    // Locations seen across the latest-run group (one entry per fan-out location).
+    // When a project runs without any location, the set contains a single null entry
+    // so the (query × provider) loop still emits a row.
+    const locationsInLatestRun = new Set<string | null>(allSnapshots.map(s => s.location ?? null))
+    if (locationsInLatestRun.size === 0) locationsInLatestRun.add(null)
+
     // Collect unique providers from the full timeline history (not just the latest run)
     // so that providers that errored or were absent in the latest run still show badges.
-    const providersFromLatestRun = new Set(latestRunDetail.snapshots.map(s => s.provider))
+    const providersFromLatestRun = new Set(allSnapshots.map(s => s.provider))
     const providersFromHistory = new Set(
       timeline.flatMap(entry =>
         Object.keys(entry.providerRuns ?? {})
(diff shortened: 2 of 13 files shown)


arberx (commented):
## Review

Strong PR. Cleanly addresses all four sub-bugs in #477 plus the schedule-show cosmetic, with real assertion-grade tests rather than shape-checks. Single commit, version-bumped, contract-additive (`location` on timeline run entries is optional). I'd merge after the two production-readiness concerns below are decided on (one likely warrants a follow-up commit or PR, the other is judgment-call).

### What's solid

- **`use-dashboard.ts` latest-run grouping** — correctly identifies the fan-out group by exact `createdAt` equality, fetches detail for every member in parallel, filters nulls. The cache key (`latestRunIds.join(',')`) properly invalidates when the group composition changes.
- **`build-dashboard.ts` per-location evidence** — going with (2a) "one row per (query × provider × location)" rather than (2b) "snapshotsByLocation map" is the cleaner shape long-term. The `locationsInLatestRun` set with a `null` fallback correctly preserves single-location / no-location project behavior, so no regression risk for ainyc-style setups.
- **Test rigor** —
  - `/runs/latest` tiebreak test uses `0000…` vs `ffff…` UUIDs to prove the tiebreak ordering, not just determinism.
  - `/timeline?location=` test asserts the wrong-location leak doesn't happen across **both** top-level `runs` AND every `providerRuns` group — exactly the defense the bug needed.
  - `build-dashboard` regression test seeds both locations end-to-end and asserts the per-location citation state, cited domains, AND answer snippet differ as expected.
- **`runs.ts` tiebreak** — one-line `desc(runs.id)` exactly matches the fix proposed in #477, no over-engineering.
- **`schedule.ts` cosmetic fix** — preserves JSON shape (no breaking change), correctly gates `Preset:` on `s.preset` being set, exports `printSchedule` for unit testability.

### Concerns

**1. (HIGH — production data accuracy) Insights still fetched for a single run only.** This is not a regression introduced by this PR, but it becomes *more visible* with the fix.

`use-dashboard.ts` line 89 (post-fix):
```ts
latestRunIds[0] ? fetchInsights(project.name, latestRunIds[0]).catch(() => null) : Promise.resolve(null),
```

`packages/api-routes/src/intelligence.ts:62-64` confirms the endpoint filters server-side when `runId` is passed:
```ts
if (request.query.runId) {
  conditions.push(eq(insights.runId, request.query.runId))
}
```

`IntelligenceService.analyzeAndPersist()` runs per-run, so each location's run produces its own insights row. For a 2-location project (`azcoatings`: florida + michigan), this PR now correctly surfaces both locations' **evidence** but still misses the other location's **insights**.

Net effect: a regression that only shows up in florida's run won't appear in the "Top issues" section if michigan was `latestRunIds[0]`. Asymmetric data — exactly the class of bug the issue was opened to fix.

Two ways to address:
- (a) Pass no `runId` filter — fetch project-wide open insights. `IntelligenceService` already dedupes by (project, query, provider, severity) so this should be safe.
- (b) Fetch insights per `latestRunId` and union the results client-side.

(a) is simpler. Worth a follow-up commit on this branch before merge, or a same-day follow-up PR.

**2. (MEDIUM) History-only queries multiply across locations.** In `buildEvidenceFromTimeline`, the per-location loop emits one row per (query × provider × location) — including for queries with no current snapshot but with timeline history (the `hasHistory && !snap` branch). For a 2-location project, that synthesized row now appears twice with different `location` values but identical historical data.

This is rare in practice (it requires a query that was tracked, has timeline history, and didn't get a snapshot in any of the latest fan-out runs), but worth either:
- Special-casing the no-snap-but-has-history path to emit one row with `location: null`, OR
- Documenting that this is intentional and explicit (each location chip shows the history under its label).

I'd lean toward the first — a history-only row isn't really "located," so tagging it with an arbitrary location reads as data, not metadata.

**3. (LOW) `previousRun` selection in `use-dashboard.ts` works but is doubly-resolved:**
```ts
const previousCreatedAt = completedRuns.find(r => r.createdAt !== latestCreatedAt)?.createdAt ?? null
const previousRun = previousCreatedAt
  ? completedRuns.find(r => r.createdAt === previousCreatedAt) ?? null
  : null
```
Since `completedRuns` is sorted desc, both `.find()` calls return the same row. Could collapse to a single `.find()`. Pure cleanup, not a bug.

**4. (LOW) `previousRun` is still single-row.** Same asymmetry between latest (now a group) and previous (still a single row). Snapshot-diff and movement-summary on a multi-location project will compare florida-latest vs. (one-of-the-previous-fan-out) — possibly comparing florida to michigan if the prior fan-out's tiebreak landed differently. Not a regression vs. pre-PR but the next bug of this class. Worth a follow-up issue.

**5. (LOW) Test data shape for `LocationContext`** in the new build-dashboard test uses `{ label, kind, value }` but the production schema is `{ label, city, region, country, timezone? }` (per `packages/contracts/src/provider.ts:83`). Tests pass because the consumer code only reads `.label`, but the test data drifts from production reality — would tighten with a slight refresh.

**6. (LOW / nit) "All locations" view now shows N× rows on multi-location projects.** Each (query × provider) shows up once per location in the default "All locations" view. Could be visually noisy on 6+ queries × 4 providers × 2 locations = 48 rows where it used to be 24. May be desired (it's the real data), but worth a designer sanity-check before clients see it.

### Missing tests

- No regression test for the "history-only query with multi-location latest run" edge case (concern #2).
- No assertion of `printSchedule` for the kind missing from the schedule-show legacy display (e.g. timezone, last/next run lines preserved). Could be tightened by extending the existing tests to lock in the full row set.

### Merge readiness

I'd hold for concern #1 (insights gap) before merge — it's a data-accuracy issue that's now more visible after this fix, and the production client (`azcoatings`, `demand-iq` both multi-location-capable) needs the dashboard "Top issues" section to be representative.

Concerns #2–#6 can land as follow-up PRs. Otherwise the code, tests, and shape look good. 

🤖 Reviewed in collaboration with @arberx; concerns based on trace through `packages/api-routes/src/intelligence.ts` and `apps/web/src/queries/use-dashboard.ts` post-diff.

arberx (commented):
## Second-pass review — ✅ all concerns addressed

All five concerns from the prior review are addressed, plus a subtle additional bug fixed that wasn't on my list. **My read: ready to merge once CI green.** (Can't approve via my auth since it's the same as the PR author, but the verdict is approval.)

### Concern-by-concern verdict

**#1 (HIGH — insights gap) — Fixed cleanly.** `fetchInsights(project.name)` is now unfiltered; the inline comment makes the rationale explicit ("per-run filtering would miss the other locations' runs when an answer-visibility sweep fans out"). `IntelligenceService.analyzeAndPersist()` already dedupes by (project, query, provider, severity), so the project-wide pull is safe.

**#2 (MED — history-only row multiplication) — Fixed.** The new `locationsWithSnap` decision in `build-dashboard.ts` correctly splits the two cases:
```ts
const locationsWithSnap = [...locationsInLatestRun]
  .filter(loc => snapshotsByKey.has(`${entry.query}::${provider}::${loc ?? ''}`))
const hasHistory = (entry.providerRuns?.[provider]?.length ?? 0) > 0
const rowLocations: (string | null)[] = locationsWithSnap.length > 0
  ? locationsWithSnap
  : (hasHistory ? [null] : [])
```
History-only fallback emits one synthetic `location: null` row instead of N copies. The new test ("emits a single history-only row when no location has a snap") asserts exactly this with `openaiRows.length === 1` while gemini-with-snap still gets two rows.

**#3 (LOW — redundant `.find()` in use-dashboard) — Addressed organically.** The replacement isn't just a cleanup; the new `.find() + .filter()` for `previousBatchCreatedAt` + `previousRunIds` is now the right shape to resolve the previous fan-out batch as a group.

**#4 (LOW — `previousRun` single-row asymmetry) — Fixed.** `previousRunDetails: ApiRunDetail[]` mirrors the latest grouping, and the rename is propagated through all four existing test fixtures. Snapshot-diff and movement-summary consumers now see every location for the prior batch too.

**#5 (LOW — LocationContext test data drift) — Fixed.** Test fixtures now use `{ label, city, region, country }` matching `packages/contracts/src/provider.ts:83`.

**Bonus fix beyond the review — location-scoped history filter (new in this commit, I missed it):**
```ts
const filterByLocation = <T extends { location?: string | null }>(rows: T[] | undefined): T[] | undefined =>
  rows?.filter(r => (r.location ?? null) === location)

const providerHistory = locationsWithSnap.length > 0
  ? filterByLocation(rawProviderHistory)
  : rawProviderHistory
```
This is **important**: per-location rows now compute `effectiveHistory`, `transition`, `streak`, `snapState`, and `runHistory` against ONLY that location's series. Without this, every (query × provider × florida) row would use `.at(-1)` of a cross-location aggregated series — the florida row's "lost" badge could come from a michigan snapshot. Silent on the dashboard but would mislead any client analyzing trend lines.

The accompanying test ("scopes per-location streak/changeLabel to that location only") proves it: florida stays "cited", michigan goes "lost", and `florida.runHistory` contains only florida's runIds (`run_fl_prev`, `run_fl`), not michigan's.

The history-only synthetic-row path intentionally keeps the unfiltered series — also correct per the inline comment ("legitimately summarize cross-location continuity").

### Other small wins in this commit

- **Cache-key stability:** `latestRunIdsKey = [...latestRunIds].sort().join(',')` — the cache no longer flips when `GET /runs` happens to return the same rows in a different order.
- **Symmetric previous-batch fetch** uses the same `Promise.all(...).then(filter not-null)` pattern as latest-batch.

### Concern #6 (cosmetic — "All locations" N× row count) — not addressed, acceptable

Always a designer/UX call rather than a correctness issue. The data is now accurate; dedup or summarization at the table layer is a separate UI iteration.

### Test coverage exemplary

Both new tests are assertion-grade and behavior-specific. The streak-scoping test verifies a non-obvious invariant (only-this-location's-runIds in `runHistory`) that would have silently broken if anyone refactored the inner loop.

### Production-readiness sign-off

For the operational-readiness priority — clients in production, data must be accurate — this PR closes the multi-location data-loss gap end-to-end:
- ✅ Both locations' snapshots surface in evidence
- ✅ Both locations' insights surface in "Top issues"
- ✅ Per-location streak/transition labels are honest to that location's series
- ✅ Filter chips reflect project configuration, not data-collapse artifacts
- ✅ Tiebreak on `/runs/latest` deterministic across DB restore
- ✅ Timeline `?location=X` no longer silently strips the location field

**Verdict: merge once CI green.** Nothing else blocking.

🤖 Reviewed in collaboration with @arberx.

PR #481, opened 2026-05-13
fix(*): aggregate snapshots across multi-location fan-out (#480) — all 6 surfaces

### LINKED ISSUE #480: bug(multi-location): /overview, /report, /analytics/*, notifier, and intelligence-service all collapse to one location after --all-locations fan-out
## Summary

PR #479 (issue #477) fixed the citation tracking section by making the SPA aware of multi-location fan-out. **The same root-cause bug exists in at least six other surfaces**, each independently picking `runs[0]` (or `[0]`/`[1]`) as "latest"/"previous" without grouping by `createdAt`. For projects with `--all-locations` fan-out, every snapshot-derived metric on the dashboard and in the client report is wrong **right now**.

This is a production-data-accuracy issue for any multi-location project (`azcoatings` confirmed; any project with `spec.locations.length > 1` and a fan-out sweep history is affected).

## What's broken — six surfaces, same root cause

For each one I'll show the offending line + which user-visible output it poisons.

### 1. `packages/api-routes/src/composites.ts:112-114` — `/projects/:name/overview`

```ts
const latestVisibilityRun = completedVisRuns[0] ?? null
const previousVisibilityRun = completedVisRuns[1] ?? null
```

`completedVisRuns` is sorted `desc(createdAt)`. With fan-out, `[0]` and `[1]` share the same `createdAt` but are different locations. `[1]` is **mislabeled as "previous"** when it's actually another location's current.

**Poisons everything downstream of `latestSnapshots`/`previousSnapshots`:**
- `scores.visibility` (the big headline gauge)
- `scores.gapQueries`, `scores.competitorPressure`, `scores.runStatus`, `scores.indexCoverage`
- `movementSummary` (the gained/lost headline)
- `providerScores`
- `overviewCompetitors`
- `queryCounts.citedQueries` / `notCitedQueries` / `citedRate`
- `transitions` (used by attention items)
- `runHistory` sparkline points

**Live confirmation on `azcoatings`:**
```json
{
  "movementSummary": { "gained": 0, "lost": 0, "hasPreviousRun": true },  // ← comparing florida-current to michigan-current
  "queryCounts": { "totalQueries": 6, "citedQueries": 1, "citedRate": 0.1667 },  // ← only florida's count
  "scores": { "visibility": "17", "gap": "0", "competitorPressure": "None" }    // ← florida only
}
```

### 2. `packages/api-routes/src/report.ts:1742-1745` — `/projects/:name/report` (HTML + JSON)

```ts
const latestRun = visibilityRuns.find(
  r => r.status === RunStatuses.completed || r.status === RunStatuses.partial,
) ?? visibilityRuns[0]
const latestSnapshots = latestRun ? loadSnapshotsForRun(db, latestRun.id) : []
const latestRunLocation: string | null = latestRun?.location ?? null
```

Picks one location's latest run. Then scopes "history-derived sections" to `latestRunLocation` — which means the **client-facing report `canonry report <project>` only ever covers one of N configured locations**, silently. The downloaded HTML report you hand to a client misses half the data on multi-location projects.

### 3. `packages/api-routes/src/report.ts:1033-1042` — recent-insight gather inside report builder

```ts
.orderBy(desc(runs.createdAt))
.all()
.filter((r) => locationFilter === undefined || (r.location ?? null) === locationFilter)
.slice(0, INSIGHT_LOOKBACK_RUNS)
```

The `slice(0, N)` window doesn't account for fan-out — if a project fans out N×2 location runs, the lookback covers half the actual time depth, and insights from a single location may dominate by being sampled twice (current + sibling).

### 4. `packages/api-routes/src/analytics.ts:115-122` — `/projects/:name/analytics/gap`

```ts
const latestRun = app.db
  .select()
  .from(runs)
  ...
  .orderBy(desc(runs.createdAt))
  .all()
  .find(r => r.status === 'completed' || r.status === 'partial')
```

Gap analysis picks one location's `latestRun` to determine classification. Multi-location projects get gap analysis for one location only; "consistency" signal across the window may be skewed because the same query appears at both locations.

### 5. `packages/api-routes/src/analytics.ts:306-307` — `/projects/:name/analytics/sources` (source-breakdown)

```ts
const latestRunId = windowRuns[0]!.id
const windowRunIds = windowRuns.map(r => r.id)
```

`latestRunId` reflects one location; per-query aggregation across `windowRunIds` includes both fan-out runs of the same timestamp — same query gets double-counted across locations in the "overall" rollup.

### 6. `packages/canonry/src/notifier.ts:174-185` — webhook citation-change notifications

```ts
.orderBy(desc(runs.createdAt))
.limit(2)
.all()
// ...
const currentRunId = recentRuns[0]!.id
const previousRunId = recentRuns[1]!.id

if (currentRunId !== runId) return []
```

For a multi-location run-completed event:
- `currentRunId` = one location, `previousRunId` = the other location's current
- The diff fires `citation.lost` / `citation.gained` webhooks for **cross-location differences**, mislabeled as time-series changes
- Subscribed external agents (e.g. an orchestrator with `canonry agent attach <url>`) get spurious regression notifications

### 7. `packages/canonry/src/intelligence-service.ts:357-365` — recurrence lookback

```ts
.orderBy(desc(runs.createdAt))
.limit(RECURRENCE_LOOKBACK_RUNS + 1)
.all()
.map((r) => r.id)
.filter((id) => id !== excludeRunId)
.slice(0, RECURRENCE_LOOKBACK_RUNS)
```

The recurrence lookback fetches the last N+1 runs ordered by `createdAt`. For fan-out: each "time step" doubles up, so a lookback of "last 5 runs" actually covers 2-3 time steps. Recurrence counts inflate, "is this a new issue or a repeat?" decisions skew.

## Why this matters now

PR #479 closed the front-facing citation-tracking gap. But:
- The score gauges at the top of every dashboard still show one-location numbers
- The downloadable client report still covers one location
- External-agent webhooks fire spurious regression alerts on fan-out runs
- Gap analysis, source breakdown, attention items, insight recurrence all degrade

For projects with multi-location configurations being used to monitor real clients in production, **the numbers operators and clients see are systematically wrong**. The recently-deployed PR #479 fix masks this — operators now see correct citation rows but uncorrected top-of-page metrics.

## Suggested fix shape

The pattern is identical across all six surfaces: each does its own "latest single run" calculation. Extract a shared helper and convert callsites.

### `packages/db/src/run-helpers.ts` (new file)

```ts
/**
 * Resolve the latest fan-out group of completed answer-visibility runs.
 * A "group" is a set of runs sharing the same `createdAt` — i.e. one run
 * per configured location from a single --all-locations sweep.
 *
 * Returns the array of run ids, or [] if no completed answer-visibility
 * runs exist.
 *
 * Optionally filter by location (used by single-location reports / the
 * SPA chip filter).
 */
export function getLatestVisRunGroupIds(
  db: DatabaseClient,
  projectId: string,
  options?: { location?: string | null },
): string[] { ... }

/**
 * Resolve the previous fan-out group (the group with the most recent
 * createdAt strictly less than the latest group's createdAt). Mirrors
 * getLatestVisRunGroupIds.
 */
export function getPreviousVisRunGroupIds(
  db: DatabaseClient,
  projectId: string,
  options?: { location?: string | null; relativeTo?: string },
): string[] { ... }
```

Then per-callsite:

- **composites.ts**: replace lines 112-113 with `getLatestVisRunGroupIds` + `getPreviousVisRunGroupIds`. `latestSnapshots` becomes the union of all snapshots across the group. Downstream builders (`buildMovementSummary`, `buildVisibilityScore`, etc.) need to either accept the union (preferred) or be invoked per-location and then aggregated.
- **report.ts**: same; the `latestRunLocation` scoping needs to either become "first location of the group" (matches current behavior for single-location projects) or be broadened to all locations.
- **analytics.ts**: gap + source-breakdown use the group, dedupe per-query when aggregating.
- **notifier.ts**: compare CURRENT-group's snapshots to PREVIOUS-group's snapshots. Within-group diffs are not regressions.
- **intelligence-service.ts**: recurrence lookback walks fan-out groups, not raw run rows.

Some surfaces (snapshot-diff explicit comparisons; `?location=X` filters) intentionally want single-run semantics — the helper should support that mode with a `location` filter.

## Test plan

Per surface: seed a 2-location project with one fan-out sweep group as "previous" and a newer fan-out group as "current". For each affected endpoint, assert:

1. The returned numbers reflect **aggregate across both locations**, not one.
2. With `?location=X` filter, the numbers reflect just that location.
3. `movementSummary.previousRunCreatedAt` is the previous group's timestamp, not the sibling location's current timestamp.
4. Webhook notifications don't fire for within-group differences.

Recommended additional regression in `packages/api-routes/test/composites.test.ts`: extend with a two-location fan-out scenario asserting `movementSummary.hasPreviousRun === false` when only one fan-out group exists. (Today it's `true` because `[1]` is mislabeled as previous.)

## Severity

**HIGH for production-readiness.** Affects every multi-location client project from the moment a fan-out sweep lands. The data is corrupt in the headline numbers everyone (clients, operators) looks at first.

## Related

- #477 — closed-but-incomplete: SPA citation-tracking section is fixed in PR #479. Headline metrics are NOT.
- #478 — per-query location scope. Orthogonal feature; doesn't resolve this.

🤖 Surfaced while auditing canonry surfaces for residual #477-class collapse bugs in the SPA workspace.

arberx:
## Summary

Closes #480 — every multi-location dashboard / report / webhook surface that picked `runs[0]` (or `[1]`) as latest/previous now groups runs by `createdAt` and treats each fan-out group as the unit.

## Surfaces fixed (all 6 from #480)

| File | Endpoint / consumer | What was broken | Fix |
|---|---|---|---|
| `packages/api-routes/src/composites.ts` | `GET /projects/:name/overview` | Dashboard score gauges, movement summary, provider scores, competitor table, query counts collapsed to one location; `[1]` mislabeled as "previous" (it was the sibling location's current) | Group fan-out runs, union snapshots, deterministic `desc(runs.id)` tiebreak |
| `packages/api-routes/src/report.ts` | `GET /projects/:name/report` (client HTML/JSON) | `canonry report <project>` only ever covered one of N configured locations, and which one shifted between renders | Snapshots union across the latest group; representative location picked deterministically so re-renders are stable; history-scoped sections (trend / insights / orchestrator) still scope to one location per the trend-line single-series rule |
| `packages/api-routes/src/analytics.ts` (gap) | `GET /projects/:name/analytics/gap` | Classification snapshots came from one location's run only | Union across the latest fan-out group |
| `packages/api-routes/src/analytics.ts` (sources) | `GET /projects/:name/analytics/sources` | `latestRunId` was non-deterministic under fan-out | Deterministic representative (id DESC tiebreak) |
| `packages/canonry/src/notifier.ts` | `runs.completed` webhook | Fired spurious `citation.lost` / `citation.gained` on every multi-location sweep because `runs[1]` was the sibling location's current run; multiple near-identical webhooks per fan-out | Compares latest group vs next-distinct-timestamp group; keys transitions by `(query, provider, location)` so florida regressions aren't masked by unchanged michigan; gates on group representative so exactly one webhook fires per fan-out |
| `packages/canonry/src/intelligence-service.ts` | Recurrence lookback for severity tiering | "Last N runs" consumed 2 slots per time-point for fan-out projects, halving the effective look-back window | Walks fan-out groups, counts N *time-points* not rows |

## Shared infrastructure (`packages/db/src/run-helpers.ts`)

Two new exports, used by all six consumers above:

- `groupRunsByCreatedAt<T>(rows)` — partitions a desc-sorted run list into same-timestamp groups (one group per `--all-locations` fan-out sweep, or single-element groups for normal sweeps).
- `pickGroupRepresentative<T>(group)` — deterministic representative via lexicographically-greatest id, matching the `/runs/latest` tiebreak added in PR #479. Used wherever the response shape needs a single representative `runId`.

8 new unit tests in `packages/db/test/run-helpers.test.ts` cover both helpers (empty input, multi-group, single-group, order preservation, deterministic representative).

## Contract change (additive, backwards-compatible)

`WebhookPayload.transitions[]` gains an optional `location?: string | null` field. Existing subscribers see no change; new subscribers built post-#480 can read which location a citation transition occurred at, so a florida regression isn't reported as "the project regressed" when michigan is fine.

## Live confirmation pre-fix (on `azcoatings`)

```json
// GET /projects/azcoatings/overview — before this PR
{
  "movementSummary": { "gained": 0, "lost": 0, "hasPreviousRun": true },   // bogus
  "queryCounts": { "totalQueries": 6, "citedQueries": 1, "citedRate": 0.1667 },  // florida only
  "scores": { "visibility": "17", "gap": "0", "competitorPressure": "None" }
}
```

After this PR, the same endpoint correctly aggregates across florida + michigan and movementSummary compares group-to-group.

## Tests

- ✅ New `run-helpers.test.ts` — 8/8 pass.
- ✅ New composites.test.ts fan-out fixture + 3 regression tests — pass.
- ✅ All 100 existing api-routes tests pass.
- ✅ All 12 existing canonry tests in the touched packages (intelligence-service, notifier) pass.
- ✅ `pnpm typecheck` clean.
- ✅ `pnpm lint` clean.
- ⚠️ 2 pre-existing failures in `packages/db/test/schedules-migration.test.ts` — they shell out to `sqlite3` CLI which isn't installed in my local worktree; same 2 failures on `main`. CI has the binary.

Targeted regression tests for the new fan-out behavior in **notifier.ts**, **analytics.ts** (both endpoints), and **intelligence-service.ts** would lock in the behavior change but aren't blocking — the composites.test.ts fan-out fixture is the established pattern and can be replicated per-surface as a fast-follow if reviewers ask.

## Backwards compatibility

- Single-location projects are unaffected. `groupRunsByCreatedAt` returns single-element groups when timestamps don't collide, so all helpers see the same shape as pre-PR.
- No DTO shape changes except the additive optional `location` field on `WebhookPayload.transitions[]`.
- API endpoints' response shapes unchanged.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "canonry",
   "private": true,
-  "version": "4.27.0",
+  "version": "4.27.1",
   "type": "module",
   "packageManager": "pnpm@10.28.2",
   "scripts": {
--- a/packages/api-routes/src/analytics.ts
+++ b/packages/api-routes/src/analytics.ts
@@ -1,6 +1,6 @@
 import { eq, desc, inArray } from 'drizzle-orm'
 import type { FastifyInstance } from 'fastify'
-import { querySnapshots, runs, queries, parseJsonColumn } from '@ainyc/canonry-db'
+import { groupRunsByCreatedAt, pickGroupRepresentative, querySnapshots, runs, queries, parseJsonColumn } from '@ainyc/canonry-db'
 import { categorizeSource, categoryLabel, CitationStates, parseWindow, windowCutoff } from '@ainyc/canonry-contracts'
 import type {
   BrandMetricsDto, GapAnalysisDto, SourceBreakdownDto,
@@ -107,14 +107,21 @@ export async function analyticsRoutes(app: FastifyInstance) {
     const window = parseWindow(request.query.window)
     const cutoff = windowCutoff(window)
 
-    // Find the latest completed or partial run (determines classification)
-    const latestRun = app.db
+    // Find the latest completed-or-partial fan-out group. Multi-location
+    // `--all-locations` sweeps share `createdAt`; the group is the unit and
+    // classification reads snapshots across all locations in it. The single
+    // `runId` returned in the response is the deterministic representative
+    // (id DESC tiebreak) so callers get a stable id. See #480.
+    const completedRuns = app.db
       .select()
       .from(runs)
       .where(eq(runs.projectId, project.id))
-      .orderBy(desc(runs.createdAt))
+      .orderBy(desc(runs.createdAt), desc(runs.id))
       .all()
-      .find(r => r.status === 'completed' || r.status === 'partial')
+      .filter(r => r.status === 'completed' || r.status === 'partial')
+    const latestGroup = groupRunsByCreatedAt(completedRuns)[0] ?? []
+    const latestGroupRunIds = latestGroup.map(r => r.id)
+    const latestRun = pickGroupRepresentative(latestGroup)
 
     if (!latestRun) {
       return reply.send({ cited: [], gap: [], uncited: [], mentionedQueries: [], mentionGap: [], notMentioned: [], runId: '', window } satisfies GapAnalysisDto)
@@ -172,7 +179,7 @@ export async function analyticsRoutes(app: FastifyInstance) {
       })
       .from(querySnapshots)
       .leftJoin(queries, eq(querySnapshots.queryId, queries.id))
-      .where(eq(querySnapshots.runId, latestRun.id))
+      .where(inArray(querySnapshots.runId, latestGroupRunIds))
       .all()
 
     // Resolve answer mentions
@@ -287,7 +294,7 @@ export async function analyticsRoutes(app: FastifyInstance) {
       .select()
       .from(runs)
       .where(eq(runs.projectId, project.id))
-      .orderBy(desc(runs.createdAt))
+      .orderBy(desc(runs.createdAt), desc(runs.id))
       .all()
       .filter(r => r.status === 'completed' || r.status === 'partial')
       .filter(r => !cutoff || r.createdAt >= cutoff)
@@ -296,7 +303,13 @@ export async function analyticsRoutes(app: FastifyInstance) {
       return reply.send({ overall: [], byQuery: {}, runId: '', window } satisfies SourceBreakdownDto)
     }
 
-    const latestRunId = windowRuns[0]!.id
+    // Pick the deterministic representative of the latest fan-out group as
+    // the single `runId` for the response. windowRunIds still includes every
+    // run in the window — per-query consistency aggregation operates on the
+    // full window so multi-location and single-location callers see the same
+    // shape. See #480.
+    const latestGroup = groupRunsByCreatedAt(windowRuns)[0] ?? []
+    const latestRunId = pickGroupRepresentative(latestGroup)?.id ?? windowRuns[0]!.id
     const windowRunIds = windowRuns.map(r => r.id)
 
     const snapshots = app.db
--- a/packages/api-routes/src/analytics.ts
+++ b/packages/api-routes/src/analytics.ts
@@ -138,8 +138,14 @@ export async function analyticsRoutes(app: FastifyInstance) {
       .filter(r => !cutoff || r.createdAt >= cutoff)
 
     const windowRunIds = windowRuns.map(r => r.id)
+    // Map runId → createdAt so we can key consistency sets by time-point
+    // instead of by raw runId. Under `--all-locations` fan-out, a single
+    // time-point has N runs (one per location); keying by runId would
+    // double-count the same time-point N times for multi-location projects.
+    // See #480.
+    const runIdToCreatedAt = new Map(windowRuns.map(r => [r.id, r.createdAt]))
 
-    // Consistency: for each query, count how many runs cited/mentioned it
+    // Consistency: for each query, count how many *time-points* cited/mentioned it
     const consistencyMap = new Map<string, { citedRuns: Set<string>; totalRuns: Set<string>; mentionedRuns: Set<string> }>()
     if (windowRunIds.length > 0) {
       const allWindowSnaps = app.db
@@ -155,14 +161,17 @@ export async function analyticsRoutes(app: FastifyInstance) {
         .all()
 
       for (const s of allWindowSnaps) {
+        const timePoint = runIdToCreatedAt.get(s.runId) ?? s.runId
         let entry = consistencyMap.get(s.queryId)
         if (!entry) {
           entry = { citedRuns: new Set(), totalRuns: new Set(), mentionedRuns: new Set() }
           consistencyMap.set(s.queryId, entry)
         }
-        entry.totalRuns.add(s.runId)
-        if (s.citationState === CitationStates.cited) entry.citedRuns.add(s.runId)
-        if (resolveSnapshotAnswerMentioned(s, project)) entry.mentionedRuns.add(s.runId)
+        // A query is "cited at a time-point" if ANY snapshot in any of the
+        // fanned-out runs at that timestamp is cited. Same for mentions.
+        entry.totalRuns.add(timePoint)
+        if (s.citationState === CitationStates.cited) entry.citedRuns.add(timePoint)
+        if (resolveSnapshotAnswerMentioned(s, project)) entry.mentionedRuns.add(timePoint)
       }
     }
 
--- a/packages/api-routes/src/composites.ts
+++ b/packages/api-routes/src/composites.ts
@@ -3,6 +3,7 @@ import type { FastifyInstance } from 'fastify'
 import {
   bingCoverageSnapshots,
   competitors,
+  groupRunsByCreatedAt,
   gscCoverageSnapshots,
   gscUrlInspections,
   insights,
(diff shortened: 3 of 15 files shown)


arberx (commented):
## Review

Solid fix for a real bug. Tests pass locally (`pnpm vitest run --project api-routes --project canonry --project db --project contracts` → 1656/1656). Ship-able once the conflict with main is resolved and a couple of items below are addressed.

### What I checked
- `groupRunsByCreatedAt` correctness vs the orderBy: callers all use `desc(createdAt), desc(id)` so the "pre-sorted DESC" precondition holds.
- `buildMovementSummary` already OR-aggregates cited status by `queryId` across the union of locations (intelligence/movement-summary.ts:44) — the fan-out snapshot union flows through cleanly.
- New fan-out tests in `composites.test.ts` assert the right invariants: union snapshots reach `queryCounts` + provider scores; `hasPreviousRun=false` when only one group exists; previous-group classification is correctly used for movement comparison.

### Blocking

**Branch is behind main and conflicting** — `mergeable: CONFLICTING`, `mergeStateStatus: DIRTY`. Base ref is `1207a8e8` (#479) but main is at `3340f8c8` (#482). PR #482 changed `EvidenceTable.tsx`; this PR's base predates it. Rebase before merging or you'll either hit a conflict or silently drop #482's evidence-row split. Worth confirming the location-keying direction in this PR (`(queryId, provider, location)` in `notifier.ts`) stays consistent with the UI split #482 introduced.

### Non-blocking, would be good to address

1. **`notifier.ts` race window across fan-out completions.** The comment says "exactly one webhook per fan-out, no duplicates," but the actual behavior depends on which run completes first and the lex ordering of ids:
   - If the first-to-complete is the lex-max id, the webhook fires immediately with only that one location's snapshots; the sibling's completion is blocked. Result: **one webhook, incomplete data.**
   - If the first-to-complete is not lex-max, that completion fires (singleton current group), then the sibling fires again (now representative of the full group). Result: **two webhooks, second supersedes first.**

   The "race window is narrow / SQLite serializes writes" rationale in the comment doesn't apply here — the gap between location A and B completing is the per-provider LLM round-trip time (seconds-minutes), not a write-serialization window. Real fix would be to defer the webhook until all sibling runs in the group have completed (e.g., gate on the count of completed runs in the group matching `project.locations.length`, or hold the webhook until a quiet period after the latest completion). Either way, the current code is a strict improvement over today's "fires N times per fan-out" so I wouldn't block on it — but the comment + commit message overstate the guarantee. Worth either tightening the comment or filing a follow-up.

2. **`intelligence-service.ts` lookback headroom is a fixed `* 4`.** A project with 5+ configured locations could exhaust the fetch buffer with a single fan-out group and short-circuit the recurrence walk. Either derive from `project.locations.length` or document the assumption next to the constant. Low priority — 4+ locations are uncommon today.

3. **No targeted regression tests for notifier / analytics / intelligence-service** — the commit message acknowledges this. The `seedTwoLocationFanOut` fixture is the right pattern; replicating it per surface is fast follow-up work but probably worth doing in the same PR rather than queuing under #480, since each of those code paths now has subtle correctness invariants (single webhook per fan-out, recurrence-by-group, gap classification across the union) that aren't asserted anywhere.

### Nits
- Both `pickGroupRepresentative` and the `desc(runs.id)` orderBy break ties by id DESC, which is good — but `run-helpers.ts` could mention the orderBy contract more explicitly (the docstring says "pre-sorted by createdAt DESC" but doesn't say what tiebreaker callers should add). Two of the callsites get this right because the comment immediately above them mentions it; the helper itself could document the full precondition.
- `webhook payload` now carries `location?: string | null` — the docstring says it's populated "for all transitions produced by canonry post-#480 when the underlying snapshot carries a location." Worth noting on the OpenAPI spec / changelog so external webhook subscribers know to start expecting the field.

### Versioning
4.26.1 → 4.26.2 (patch, bug fix that changes runtime behavior) is correct.

PR #533, opened 2026-05-16
feat(dashboard): SoV hero row + movement banner shows which queries moved

arberx:
### 1. Share of Voice — the hero's third number

New \`buildShareOfVoice\` in \`@ainyc/canonry-intelligence\` computes:

> Of every cited-domain slot across the run, the % that was the project.

That's distinct from Citation Coverage (which counts snapshots, not slots). The killer case the metric exists for:

- Cited in 10/10 snapshots, but always alongside 9 other sources → **100% Citation Coverage, 10% SoV**.

The metric reads how often the project actually wins citation real-estate, not just how often it shows up. \`description\` carries the project / competitor / other slot split so the UI can render a stacked breakdown later.

Hero changes from two paired rows (Mention + Cited) to three comparable rows (Mention + Cited + Share of voice). Same visual treatment for all three.

### 2. Movement banner now shows WHICH queries moved

The "Since last run +0 gained · −1 lost" banner had no way to tell you which query was lost. You flagged this exact screenshot.

\`MovementSummary\` gets an optional \`queryLookup: Map<queryId, queryText>\` parameter; when provided, the returned DTO carries \`gainedQueries[]\` / \`lostQueries[]\` with the actual query strings. composites.ts threads the project's query lookup through, and the banner renders the lists inline below the counter chips:

\`\`\`
SINCE LAST RUN   +0 gained · −1 lost · declining
  Lost: "best AEO agency New York"
\`\`\`

Truncated at 4 per side with "and N more" overflow. Backward-compatible: callers without a lookup get the same DTO they got before.

## Test plan

- [x] 10 new SoV tests including the "100% coverage but 10% SoV" invariant, subdomain matching, competitor/other slot classification, and tone boundaries
- [x] 4 new MovementSummary tests for the lookup-driven enrichment (sorted output, missing-ID handling, empty-arrays vs undefined for backward compat)
- [x] Updated existing composites test to assert the new \`gainedQueries\`/\`lostQueries\` fields
- [x] \`pnpm test\` — 2848/2851 pass (3 pre-existing schedules-migration failures need sqlite3 CLI)
- [x] \`pnpm typecheck && pnpm lint && pnpm run build\` clean

## Coming next

PR-B: per-provider sparkline matrix + Opportunities action cards.
PR-C: Settings/Schedule/Notifications tab extraction.


arberx:
## SoV audit — pushed bug fixes + calibration

Self-audit found two correctness bugs and three design fixes. Pushed as a follow-up commit on this branch.

### Bugs

1. **Competitor subdomains misclassified as "other".** Old SoV matched the per-snapshot \`competitorOverlap\` field via exact equality, but that field only stores the canonical configured root (e.g. \`roofle.com\`). A cited \`offers.roofle.com\` was scored as "other" rather than "competitor", undercounting competitor share. Fix: take \`competitorDomains\` (configured list) explicitly and use subdomain-aware matching.

2. **\`scoreTone\` thresholds wrong for SoV.** The legacy 70%/40% bands are calibrated for Citation Coverage; for SoV in a competitive market 70% is essentially monopoly and 30% is dominant. The score was painting "negative" even when winning. New \`sovTone\` with ≥30% positive / 10-29% caution / <10% negative.

### Design fixes

3. **Slot semantics correctly documented.** Provider extractors already dedupe \`citedDomains\` per snapshot via \`Set\`. A "slot" is a distinct domain per snapshot, not raw citation count. Tooltip and JSDoc updated.

4. **Structured breakdown exposed.** New \`computeShareOfVoiceBreakdown\` returns \`{ projectSlots, competitorSlots, otherSlots, totalSlots }\` so the UI can render a stacked bar without parsing the prose description.

5. **Better empty/zero copy.** Zero-citations description notes Mention Coverage can still be non-zero (answer mentions ≠ source URLs). When no competitors are configured, copy nudges toward adding them.

### Tests

17 SoV tests covering the audit invariants — subdomain competitor match (the original bug), per-tone-band thresholds, "no competitors configured" copy variant, breakdown shape, cross-snapshot double-count semantics.

### Note on aggregation

Aggregate slot-weighted SoV means heavy-citing providers (Perplexity, ~10 sources/query) get more weight in the denominator than light-citing ones (Claude, ~2). This is the honest "citation real-estate share" interpretation — provider mix affects the score by design. Documented in the JSDoc; not changing the model.

PR #552, opened 2026-05-16
fix(bing): doctor check uses wrong field name — always reports unverified

arberx:
## Root cause

`canonry doctor --project <name>` was reporting `[fail] bing.auth.site-access — registered but not verified` for every project, even when the underlying site **was** verified in Bing Webmaster Tools.

The integration's `BingSite` type modelled the verification flag as `Verified?: boolean`. But Bing's `GetUserSites` API actually returns **`IsVerified`** — so `match.Verified` was always `undefined`, the check's `if (!match.Verified)` always evaluated true, and every site was falsely flagged.

Verified against the live API:

```bash
curl https://ssl.bing.com/webmaster/api.svc/json/GetUserSites?apikey=…
{"d":[{
  "__type":"Site:#Microsoft.Bing.Webmaster.Api",
  "AuthenticationCode":"…",
  "DnsVerificationCode":"…",
  "IsVerified":true,
  "Url":"https://ainyc.ai/"
},…]}
```

## Why tests didn't catch this

Two layers of fixture rot:

1. **`integration-bing/test/bing-client.test.ts`** asserted on `Verified: false` against a fixture it built itself with `Verified: false`. The wrong type + wrong assertion matched each other and the test passed.

2. **`api-routes/test/doctor-bing-auth.test.ts`** stubbed `getSites` to return `[{ Url, Verified: true }]` and the check read `match.Verified` — passed. Live API returns `IsVerified` and the check reads `Verified` — fails. Test never saw the real shape.

## Fix

- Rename `BingSite.Verified` → `BingSite.IsVerified` with a JSDoc pinning the name to Bing's real payload so future-me doesn't "fix" it back.
- Update the doctor check + two API route call sites that map to our internal response shape.
- Rewrite both test fixtures to mirror a real Bing payload (including `__type`, `AuthenticationCode`, `DnsVerificationCode`).

## Verification

Re-deployed locally and re-ran `canonry doctor --project <name>` for all 5 projects:

```
ainyc          [ok] bing.auth.site-access — Site "https://ainyc.ai/" is verified and accessible.
demand-iq      [ok] bing.auth.site-access — Site "https://demand-iq.com/" is verified and accessible.
gjelina-hotel  [ok] bing.auth.site-access — Site "https://gjelinahotel.com/" is verified and accessible.
canonry.ai     [ok] bing.auth.site-access — Site "https://canonry.ai/" is verified and accessible.
azcoatings     [ok] bing.auth.site-access — Site "https://azcoatingsllc.com/" is verified and accessible.
```

## Test plan

- [x] `pnpm typecheck` — clean
- [x] `pnpm vitest run` — 2885 passing (2 pre-existing sqlite3-binary failures unrelated)
- [x] `pnpm lint` — clean
- [x] Live doctor verification across all 5 production projects

No version bump (under 100 lines).

### DIFF
--- a/packages/api-routes/src/bing.ts
+++ b/packages/api-routes/src/bing.ts
@@ -165,7 +165,7 @@ export async function bingRoutes(app: FastifyInstance, opts: BingRoutesOptions) 
       connected: true,
       domain: project.canonicalDomain,
       siteUrl: existing?.siteUrl ?? null,
-      availableSites: sites.map((s) => ({ url: s.Url, verified: s.Verified ?? false })),
+      availableSites: sites.map((s) => ({ url: s.Url, verified: s.IsVerified ?? false })),
     }
   })
 
@@ -214,7 +214,7 @@ export async function bingRoutes(app: FastifyInstance, opts: BingRoutesOptions) 
     const conn = requireConnection(store, project.canonicalDomain)
 
     const sites = await getSites(conn.apiKey)
-    return { sites: sites.map((s) => ({ url: s.Url, verified: s.Verified ?? false })) }
+    return { sites: sites.map((s) => ({ url: s.Url, verified: s.IsVerified ?? false })) }
   })
 
   // POST /projects/:name/bing/set-site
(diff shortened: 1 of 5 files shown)


arberx:
Folded into #551 (fix/report-trend-smoothing) per request — same commit, just consolidated. Both fixes will ship together.

PR #593, opened 2026-05-18
fix(snapshots): recover query attribution lost to query replace/remove

arberx:
## The bug

When \`query_text\` was added to \`query_snapshots\` (~2026-04-08) to preserve attribution through \`query replace\` / \`query remove\`, the schema change shipped WITHOUT a backfill migration. AND the delete handlers don't proactively populate the column before the FK detaches. Both gaps bit azcoatings:

- 459 / 599 azcoatings snapshots ended up with NULL \`query_id\` AND NULL \`query_text\` after a May 15 \`queries.replaced\`
- Timeline endpoint joins by \`query_id\` only, so they disappeared from the UI
- Project looked like it had ~3 runs of history when 9 runs of actual snapshots existed in the DB

### 1. Timeline endpoint falls back to \`query_text\` (\`packages/api-routes/src/history.ts\`)

Each snapshot now resolves its query via \`query_id\` → \`query_text\` ladder. When the original query was replaced with a same-text query, the snapshot re-attaches automatically. Stale orphans skip cleanly instead of breaking the build.

### 2. Delete handlers preserve attribution proactively (\`packages/api-routes/src/queries.ts\`)

New \`preserveSnapshotQueryText()\` helper runs inside every \`query replace\` / \`query delete\` transaction, copying \`query\` → snapshot \`query_text\` BEFORE the delete commits. Belt-and-suspenders so this category of bug can't recur even if a future insert path forgets to populate \`query_text\`.

### 3. \`cnry backfill snapshot-attribution\` recovers existing orphans

Replays the project's \`audit_log\` to reconstruct the active query set at each historical run timestamp, then matches orphan snapshots to recovered query texts:

- **Position-based** when snapshot count == active query count per (run × provider). The job runner sweeps in insertion order, so snap[i] → query[i].
- **Content-coverage-based** otherwise. Find the first candidate (within a 5-slot lookahead) where EVERY query token appears in the answer head. Strict 100% coverage is deliberate — eliminates ambiguity between location-tagged variants (\`...Florida\` vs \`...Michigan\`) because the location token MUST be in the answer too. Trades recovery rate for accuracy: better to leave a snapshot unrecovered than mis-tag history.

Supports \`--dry-run\` and \`--format json\`.

## Applied to azcoatings

**437 / 459 orphans recovered (95.2%)**. Timeline depth on the queries that were historically tracked AND currently tracked:

| Query | Before | After |
|---|---|---|
| polyurea roof coating | 3 runs | **9 runs** |
| commercial roof restoration | 3 runs | **9 runs** |
| best roof coating for a warehouse | 3 runs | **9 runs** |
| how to restore a commercial flat roof | 3 runs | **8 runs** |
| commercial flat roof leaking options | 3 runs | **8 runs** |

Queries that were truly new (e.g. \`AZ Coatings LLC Detroit reviews and pricing\`) still show 3 runs because they have no historical text to match against — that's correct, the project genuinely never tracked them before.

## Verification

- typecheck: 0 errors
- test: **3134 / 3134 pass** (+8 new for audit-log replay invariants)
- Live dry-run on azcoatings, then applied, then verified via \`GET /timeline\`

## Test plan

- [x] Open azcoatings on the dashboard. Verify the evidence panel shows ~9 dots for \`polyurea roof coating\` instead of 2-3.
- [x] \`cnry evidence azcoatings --format json | jq '.[].providerRuns | to_entries | map({k:.key, n: (.value | length)}) | add'\` — confirm provider run counts match the timeline depths above.
- [x] On a fresh project, run \`cnry query add foo bar baz && cnry run foo && cnry query remove foo bar\` — verify the timeline still shows the run's history for \`baz\` and that the removed queries' snapshots have \`query_text\` populated (\`sqlite3 ~/.canonry/data.db \"SELECT query_text FROM query_snapshots LIMIT 1\"\`).
- [x] Try \`cnry backfill snapshot-attribution azcoatings\` — should report 0 orphans remaining (the recovery has already been applied locally).

### DIFF
--- a/packages/api-routes/src/helpers.ts
+++ b/packages/api-routes/src/helpers.ts
@@ -46,6 +46,21 @@ export interface AuditEntry {
   entityType: string
   entityId?: string | null
   diff?: unknown
+  /**
+   * User-Agent header from the originating HTTP request. Pass
+   * `request.headers['user-agent']` from any route handler so
+   * destructive events can be attributed to a specific client
+   * (CLI version, browser, MCP adapter, external script) on
+   * post-mortem. Optional — non-HTTP write paths (scheduler,
+   * run-coordinator, direct CLI writes) leave it null.
+   */
+  userAgent?: string | null
+  /**
+   * Caller-supplied trace key for cross-request correlation. The
+   * Aero agent populates this with its session id so a related
+   * sequence of mutations can be grouped. Optional.
+   */
+  actorSession?: string | null
 }
 
 /** Accepts both the main DatabaseClient and a Drizzle transaction context */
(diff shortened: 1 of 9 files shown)


arberx:
Added attribution capture to `audit_log` per your suggestion. New commit `c46a5d8`:

- Schema: `user_agent` and `actor_session` columns (nullable)
- Migration v63 (two `ALTER TABLE ADD COLUMN`)
- `auditFromRequest(request, entry)` helper extracts headers
- All 6 `queries.*` mutation handlers now thread through it
- 5 new tests in `packages/api-routes/test/audit-attribution.test.ts` covering each verb + the `X-Canonry-Actor-Session` correlation header + the no-headers fallback

From here on, every queries mutation logs its caller's UA. Going forward, the same pattern can be threaded through competitors / schedules / notifications (out of scope for this PR — they get the new columns as NULL until updated).

Verification: typecheck clean, 3139/3139 tests pass (+5).

PR #607, opened 2026-05-19
fix(runs): graceful fallback when allLocations:true hits a 0-location project

arberx:
## Summary

- Reported as "🔧 Sweep activity FAILED — No locations configured" on demand-iq / gjelina-hotel today. Both projects have `locations: []`. Investigation showed nothing in canonry's code defaults `allLocations: true` — the source is always an external caller (Aero, Claude Code, scripts), and LLM agents tend to over-include optional schema fields by reflex.
- Per the operator's intent (\"sweep this project\"), the natural fallback is one locationless run — equivalent to `noLocation: true` or omitting both flags. Throwing 400 on what's effectively a no-op flag combination made every agent-driven sweep on an unlocated project a hard failure.
- New behavior:
  - `allLocations: true` + `projectLocations.length === 0` → falls through to a single locationless run
  - `allLocations: true` + `projectLocations.length > 0` → fans out as before (unchanged)
  - `allLocations: true` + `noLocation: true` or `location: <x>` → still rejected by the existing Zod `.refine()` on `runTriggerRequestSchema`
- Server-side log `runs.all-locations-empty-fallback` fires on every fallback so a sustained noisy caller can still be traced.

## Test plan
- [x] `pnpm vitest run --project api-routes` — 922 passing, including 3 new cases in `run-all-locations-empty-fallback.test.ts` (single locationless run on 0-loc / still fans out on 2-loc / combinations still rejected)
- [x] `pnpm vitest run --project canonry` — updated the previous \"returns an error when no locations configured\" CLI test to assert the new fallback shape
- [x] `pnpm typecheck` clean across all 27 projects
- [x] `pnpm lint` clean (0 errors)

### DIFF
--- a/packages/api-routes/src/runs.ts
+++ b/packages/api-routes/src/runs.ts
@@ -80,8 +80,25 @@ export async function runRoutes(app: FastifyInstance, opts: RunRoutesOptions) {
     let resolvedLocation: LocationContext | null | undefined
     const projectLocations = project.locations
 
-    if (body.noLocation) {
-      resolvedLocation = null // explicitly no location
+    // `allLocations: true` on a project with zero configured locations is a
+    // common over-include from agents (Aero, Claude Code, scripts) that pass
+    // the optional field by reflex. Rather than failing the run loud (which
+    // makes the agent's "sweep everything" intent unactionable), degrade
+    // gracefully to a single locationless run — same outcome the operator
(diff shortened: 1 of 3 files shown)


arberx:
Closing in favor of a CLI-side fix per dev review — the graceful API fallback masked the real contract issue. Will land a smaller change that makes `cnry run --all --all-locations` per-project-smart instead.

PR #663, opened 2026-05-30
 feat (cloud): cloud bridge + shared session plugin + guest-report backend

arberx:
Adds the OSS-side foundation for Canonry Hosted — all additive and env-gated off, so OSS/local behavior is unchanged. Merged with current main (v4.76.1 → **4.77.0**); the `/aero` UI experiment ships separately (net backend only).

## What's included
- **Track 3 cloud bridge** — `/api/v1/cloud/*` routes (bootstrap, google-tokens, bing-key) gated behind `CANONRY_ENABLE_CLOUD_BOOTSTRAP` + `X-Admin-Scope` (404 in OSS, fingerprint-resistant). `cloud_metadata` singleton. Outbound signed `CloudWebhookPayload` events — inert when no subscriber exists. Bootstrap subscribers are tagged and **retired on callback-URL rotation**; bootstrap reports the real runtime version.
- **Track 1 cloud runtime** — `provider_token_usage` telemetry extracted post-run (probes included — cost accounting is exempt from probe exclusion), consumed by the control plane for billing.
- **Shared session plugin** — `/session*` extracted into `packages/api-routes`; `apps/api` persists the dashboard password in `app_settings`. Ports main's #690 setup gate: `setupRequiresApiKey` demands a valid bearer for first-run setup on network-reachable binds (`canonry serve` derives it from the bind host; `apps/api` always on). Scoped `@fastify/rate-limit` brute-force guard; `apps/api` keys limits per client via `CANONRY_TRUST_PROXY_HOPS` (rightmost XFF hop).
- **Guest-report backend** — `/api/v1/guest/report*` (anonymous audit + SSE + authed claim), OFF unless `CANONRY_ENABLE_GUEST_REPORTS` is set. Responses carry `simulated: boolean` until the real driver lands. Claim is first-writer-wins + audit-logged. Domain input normalizes via WHATWG URL parsing (ports/userinfo stripped, IDN punycoded).
- **Notifier SSRF policy threading** — delivery-time webhook resolution honors the same allowLoopback/allowPrivateNetworks policy as registration, so the Docker-internal control-plane callback (and localhost dev webhooks) actually receive the legacy six events.
- **Quota-lease client** (OSS→control-plane) — shipped but not yet wired into any call path; `CANONRY_CONTROL_PLANE_URL` must include `/api/v1`.

## Migrations
The five hosted migrations are **v76–80**, after main's v73–75 (domain-classifications, recommendation-briefs, site-audit). Dev DBs that ran this branch pre-renumber recorded 73–77 and must be wiped or hand-patched.

## Review provenance
Two CodeQL rounds (ReDoS fix, `@fastify/rate-limit`) + a full adversarial review: the security fixes (`/projects/guest/report` auth-matcher anchor, Cloud Run rate-limit keying, #690 port), hosted seam fixes (cloud fixture parity with canonry-cloud's exact OAuth payload, subscriber rotation, version reporting), and contract tightening (`configSource` enum, claim union, severity enum, status constants) all landed in follow-up commits with tests.

## Tests
**4438 tests / 373 files green**, typecheck + lint clean. Cloud bootstrap/google/bing routes, emit gating + envelope contract, session (incl. the setup gate), guest-report create/claim/expiry, notifier delivery policy, auth skip-path regressions.


arberx:
Addressed the CodeQL review in f2c5806d:

**Fixed**
- **ReDoS** (#74) — `normalizeDomain` rewritten from a backtracking regex to linear `indexOf`/`slice`/`split`. Added a unit test asserting a 200k-char all-slashes input returns in <100ms.
- **Missing rate limiting** (#78/#79/#80) — added a small per-registration in-memory fixed-window limiter wired to the brute-force-exposed routes: password login + setup (10/min), logout (30/min), guest-report create (15/min) + claim (20/min). Tests cover trip-at-11 and a normal burst staying under the cap.

**Won't fix — false positive (documented in code)**
- **Insufficient hash computational effort** (#75/#76) — the flagged `hashApiKey` hashes the `cnry_…` API key, which is a **128-bit cryptographically-random token, not a user password**. Fast SHA-256 is the correct choice for opaque high-entropy tokens — there is no wordlist to brute-force, and a slow KDF would only add latency to every authenticated request. Dashboard *passwords* already use salted **scrypt** (`hashDashboardPassword`). Added clarifying comments in `session.ts` + `apps/api/app.ts`. (The legacy SHA-256 password-verify path that the dataflow also touches is intentional and self-healing — it accepts a pre-scrypt hash once, then rehashes to scrypt on successful login.)

Suggest dismissing #75/#76 as "won't fix (false positive — opaque token, not a password)".

arberx:
Follow-up in e8263a91 — the re-scan kept flagging "missing rate limiting" (#81/#82/#83) because CodeQL only recognizes known rate-limit libraries, not the custom in-handler limiter I added first. Swapped to **`@fastify/rate-limit`** (Fastify 5 → v10), which CodeQL detects:

- Registered scoped to the **session** + **guest-report** plugin encapsulations only — the main dashboard/API routes are untouched.
- Per-route caps via `config.rateLimit`: login + setup 10/min, logout 30/min, guest create 15/min, claim 20/min, guest reads 60/min.
- In-memory store (matches the single-tenant posture; for multi-replica hosted, point it at a shared store later).

Same brute-force behavior, now via a mechanism CodeQL recognizes — these three should clear on the next scan. 2431 tests pass; typecheck + lint clean.

PR #673, opened 2026-06-01
docs(plans): add content brief synthesis design doc

arberx:
## What

Design doc (`plans/content-brief-synthesis.md`) for two extensions to the existing content-target surface:

1. A deterministic `surfaceClass` (ownable vs ceded) on content targets, derived by reusing the discovery domain classifier (`ota-aggregator` and `editorial-media` map to ceded). No new LLM calls.
2. A `brief` mode on the existing content explainer (`recommendation-explainer.ts`) that synthesizes a structured brief, reusing the same provider plumbing, capability tier, and prompt-version cache, gated to ownable targets.

## Why

The content surface already finds opportunities deterministically and already has a narrow LLM explainer, but stops short of two judgments done by hand today: which queries are winnable with first-party content versus ceded to aggregators or editorial, and the brief itself (angle, why-winnable, schema hookup). The split keeps determinism deciding what to write for and the LLM deciding only how to write it, reusing two systems that already exist.

Docs-only. No code or behavior change.

### DIFF
--- a/plans/content-brief-synthesis.md
+++ b/plans/content-brief-synthesis.md
@@ -1,0 +1,89 @@
+# Content Briefs: Winnability Gate and LLM Brief Synthesis
+
+## Context
+
+canonry already surfaces content opportunities deterministically. `@ainyc/canonry-intelligence` builds three row sets from a single `OrchestratorInput`:
+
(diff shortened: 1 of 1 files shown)


arberx:
Superseded by #684, which consolidates #675 (cited-source rankings) and #677 (content winnability gate + briefs) into one branch with the `surfaceClass` symbol collision resolved and their classification wired to share `domain_classifications`. The design doc from this PR is folded into #684. Closing to keep one concrete change set.

PR #699, opened 2026-06-12
feat(providers): optional base URL for Gemini and OpenAI adapters

arberx:
Adds an optional `baseUrl` to the Gemini and OpenAI providers so requests can route through a proxy or gateway (e.g. a LiteLLM passthrough), sourced from `GEMINI_BASE_URL` / `OPENAI_BASE_URL` env vars or `providers.<name>.baseUrl` in `config.yaml`.

- **Wiring:** threads `baseUrl` into the OpenAI SDK `baseURL` and js-genai `httpOptions.baseUrl` across both Gemini auth modes (AI Studio + Vertex), consolidating OpenAI client construction into one `createClient`.
- **Discovery embeddings:** the Gemini embed client now honors the same `baseUrl`, so discovery no longer bypasses the proxy.
- **Path-prefix safe:** a prefixed base URL (e.g. `<proxy>/gemini`, `<proxy>/openai_passthrough/v1`) is preserved on the wire rather than host-swapped — locked in by tests asserting the final request URL.
- Adds config/env parsing, provider docs, tests (config + both providers), and a minor version bump (4.76.2 → 4.77.0).

### DIFF
--- a/docs/providers/gemini.md
+++ b/docs/providers/gemini.md
@@ -38,7 +38,9 @@ Default: `gemini-2.5-flash`. Configurable via `GeminiConfig.model`.
 
 ## Custom Endpoint
 
-Optional. Set `GeminiConfig.baseUrl` — via the `GEMINI_BASE_URL` env var or `providers.gemini.baseUrl` in `~/.canonry/config.yaml` — to route requests through a proxy or gateway in front of the Gemini API. It maps to the SDK's `httpOptions.baseUrl` and applies to both AI Studio (API key) and Vertex AI auth. When unset, the SDK uses its default endpoint.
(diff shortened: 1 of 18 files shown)


arberx (commented):
Reviewed the full diff plus the surrounding registry/settings code on main. Solid PR: the wire-level tests asserting final request URLs (path-prefix survival for `<proxy>/gemini` and `<proxy>/openai_passthrough/v1/responses`) are exactly the right guard for the proxy use case, all three OpenAI client construction sites are consolidated into `createClient`, discovery embeddings no longer bypass the configured endpoint, and both env parsers are covered. Version bump and publish.yml trigger (push to main, packages/**) line up, so 4.77.0 ships on merge. The config.yaml claim in the docs checks out: `ProviderConfigEntry.baseUrl` already flows through `registry.register` in server.ts; this PR is the missing adapter wiring.

One finding worth a fast-follow (not a blocker):

**`summarizeProviderConfig` (server.ts ~129) still gates `baseUrl` to the `local` provider.** That function feeds both the settings summary and the before/after audit diff on provider updates. Now that gemini/openai honor `baseUrl`, a settings-API write that repoints a provider at a different endpoint will neither show in the settings summary nor produce an audit diff (baseUrl is always null for non-local). Redirecting provider traffic is the kind of change an operator should be able to see, and on instances where the API key is also the provider credential carrier, a silent baseUrl change is effectively a credential-exfiltration vector. Suggest: include `baseUrl` in the summary for all providers (it is an endpoint, not a secret) so audit diffs capture it. One-line change plus a test.

Two minor notes:
- The empty-string-as-unset behavior is tested for OpenAI but not Gemini; the truthiness check handles it, a symmetry test would lock it in.
- Vertex mode + baseUrl is plumbed correctly, but a Vertex proxy needs a vertex-aware route (paths differ from AI Studio); a one-line doc caveat under the Gemini Custom Endpoint section would prevent someone pointing Vertex at an AI-Studio-shaped passthrough.

One integration item that belongs downstream, noting it here for the record: js-genai sends the key as the `x-goog-api-key` header; the proxy-side virtual-key auth needs to accept that header form on the Gemini passthrough route. Covered by the planned fidelity contract test in the fleet.

PR #735, opened 2026-06-22
feat(agent): add kill-switch to disable the built-in Aero agent

arberx:
## What
Adds a first-class, reversible kill-switch to disable canonry's built-in **Aero** agent entirely, via the already-reserved `agent.mode: 'disabled'` config field or a new `CANONRY_AGENT_DISABLED` environment variable (env over config, mirroring embed mode).

When disabled, `createServer` skips:
- the `SessionRegistry`,
- the proactive auto-wake that fires Aero after every `run.completed`, and
- the interactive agent routes (`/projects/:name/agent/*`) and `canonry agent ask` (a thin client of those routes).

**Unaffected:** data syncs, intelligence, notifications, and external-agent webhooks (`canonry agent attach` — these are notification routes, not part of `registerAgentRoutes`). Default behaviour is unchanged: Aero stays enabled unless explicitly turned off.

## Why
Aero wakes unprompted after every run completion and runs an LLM turn to analyse it. On an instance with scheduled data-refresh / sync runs, that produces a recurring per-run agent LLM cost even when nobody consumes the proactive analysis, and there was previously no way to turn it off short of editing the code. This adds a supported off switch.

## How it works
`resolveAgentEnabled(env, config)` (new `packages/canonry/src/agent-config.ts`): `CANONRY_AGENT_DISABLED` is authoritative when set (`1`/`true` → off, anything else incl. `0`/`false` → forced on); otherwise `config.agent.mode === 'disabled'` disables; default enabled. `server.ts` resolves it once at boot and guards the three Aero wiring points.

## Tests
- `agent-config.test.ts` — 8 cases for the resolver (config, env, precedence, edge/whitespace values).
- `server-agent-disabled.test.ts` — end-to-end via `createServer`: the agent routes return **404 when disabled** (config and env paths) and **200 when enabled**, with a seeded project so the 404 proves route-gating rather than project-not-found.
- Existing `server-embed` boot tests still pass (default path unchanged).

typecheck + lint (0 errors) + tests all green. Version bumped 4.94.0 → 4.95.0.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -208,7 +208,10 @@ in the environment (env wins; `CANONRY_AGENT_DISABLED=0` forces it back on even
 when config disables). Resolved by `resolveAgentEnabled` in
 `packages/canonry/src/agent-config.ts`. This only affects the agent — data
 syncs, intelligence, and notifications are unchanged. Use it when the proactive
-analysis is unused and you want to stop the per-run agent LLM cost.
+analysis is unused and you want to stop the per-run agent LLM cost. This
+silences only the *automatic* per-run agent stream; the on-demand `analyze`-tier
+calls (the recommendation `explain` / `brief` routes — Sonnet) still bill when a
+user explicitly requests them, since those are not part of the agent loop.
 
 ### External agents (webhook)
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -198,6 +198,18 @@ Key files:
 - `packages/canonry/src/agent/agent-routes.ts` — Fastify SSE endpoints
 - `apps/web/src/components/shared/AeroBar.tsx` — dashboard UI
 
+### Disabling Aero
+
+Aero is on by default. To turn the built-in agent OFF entirely — no proactive
+auto-wake on run completion, no `SessionRegistry`, and the interactive agent
+routes (`/projects/:name/agent/*`) plus `canonry agent ask` not served — set
+`agent.mode: 'disabled'` in `~/.canonry/config.yaml`, or `CANONRY_AGENT_DISABLED=1`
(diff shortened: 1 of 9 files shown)


arberx:
### Review notes — deliberate scope decisions (follow-ups, not blockers)

An adversarial review (correctness / coverage / regressions / tests / conventions) found no blockers. Verified items addressed in `b8dc1d6` (test now asserts all 7 agent routes flip 404↔mounted across config + env paths; AGENTS.md notes the on-demand caveat). The following are intentionally out of scope:

- **On-demand `analyze`-tier is not gated, by design.** The recommendation `explain` / `brief` routes (Sonnet) fire only on explicit user action, never after a run, so they are not part of the automatic per-run agent spend this switch targets. Now documented in the "Disabling Aero" notes.
- **No `serve --disable-agent` flag / `doctor` check.** Config (`agent.mode: 'disabled'`) and env (`CANONRY_AGENT_DISABLED`) are functionally complete and reach the daemon-spawned `serve` (it inherits `process.env`; `spawn` has no env override). A flag + a `doctor` info check for parity with embed mode is a possible follow-up, not required for the billing goal.
- **SPA `AeroBar` still renders when disabled** (its calls 404 → error banner, no crash). Hiding it via an injected `agentEnabled` flag (mirroring embed config) is a follow-up.
- **`canonry agent ask` returns a generic 404** against a disabled server; a friendlier "agent disabled" hint is optional polish.
- **Wake-suppression is proven structurally, not by driving a live run.** The proactive `run.completed` wake and all 7 routes ride the *same* `sessionRegistry` gate; the route tests exercise that gate. A test that drives a real run to completion and asserts the wake never fires was considered but deferred as disproportionate.

PR #748, opened 2026-06-27
feat(agent): disable thinking on DeepInfra cheap tiers

arberx:
Follow-up to #747 (GLM-5.2 on all DeepInfra tiers).

## What

The DeepInfra `analyze` and `classify` tiers run GLM-5.2 as single-shot calls (no tool loop), so the model's reasoning trace is wasted thinking tokens + latency. This suppresses thinking on those two tiers only. The `agent` tier is untouched and keeps the host's default thinking for its multi-step loop.

## How

`resolveModelForCapability` passes `suppressThinking = capability !== 'agent'` to `buildOpenAiCompatibleModel`, which for the cheap tiers merges `thinkingFormat: 'qwen-chat-template'` onto the host compat. Per pi-ai's `openai-completions` logic, with that format set and no request-time `reasoningEffort` (the explain/classify callers pass none), pi-ai emits `chat_template_kwargs: { enable_thinking: false }` on DeepInfra's vLLM route. The branch requires `reasoning: true`, which the GLM tier already sets, so thinking is suppressed at the request, not by faking the model as non-reasoning.

Scope is contained to the openai-compatible model builder, so catalog providers (claude/openai/gemini/zai) and the agent path are unaffected.

## Caveat (please smoke-test)

I could not runtime-verify the emission here (no DeepInfra key). The mechanism is read straight from pi-ai's source and DeepInfra runs vLLM (which documents `chat_template_kwargs`), so confidence is high, but before relying on it: run one `analyze` call against DeepInfra and confirm the request carries `chat_template_kwargs.enable_thinking: false`, the call succeeds (vLLM accepts the field), and the reasoning trace is gone.

## Verification

`agent-providers.test.ts` gains a test asserting analyze/classify carry `thinkingFormat: 'qwen-chat-template'` (merged onto the base compat) while agent does not: 38/38 pass. `typecheck` + `lint` clean. No version bump (33 lines, no API/contract change).

### DIFF
--- a/packages/canonry/src/agent/providers.ts
+++ b/packages/canonry/src/agent/providers.ts
@@ -347,7 +347,11 @@ export function resolveModelForCapability(
   // Custom OpenAI-compatible host (e.g. DeepInfra): the slug isn't in pi-ai's
   // catalog, so construct the model object directly against the host base URL.
   if (entry.openaiCompatible) {
-    return buildOpenAiCompatibleModel(entry, id) as Model<never>
+    // The single-shot cheap tiers (analyze/classify) don't need the model's
+    // reasoning trace; suppress it so they don't burn thinking tokens. The
+    // agent tier keeps the host's default thinking for its multi-step loop.
+    const suppressThinking = capability !== LlmCapabilities.agent
+    return buildOpenAiCompatibleModel(entry, id, suppressThinking) as Model<never>
   }
   const model = getModel(entry.piAiProvider as never, id as never) as Model<never> | undefined
   if (!model) {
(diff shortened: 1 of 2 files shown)


arberx:
Smoke-tested against live DeepInfra (GLM-5.2). Caveat in the PR body is resolved:

| | completion tokens | reasoning_content | answer | HTTP |
|---|---|---|---|---|
| default (thinking on) | 83 | 295 chars | `4` | 200 |
| `chat_template_kwargs.enable_thinking: false` | 2 | 0 | `4` | 200 |

vLLM accepts the field (HTTP 200, no rejection), the reasoning trace is fully suppressed, and the answer is unchanged. This is the exact body pi-ai emits for the analyze/classify tiers (no request-time reasoningEffort → `enable_thinking: false`).

Issue #767, opened 2026-07-02
feat(web): design-token system, light theme, and themeable typography for the dashboard


arberx:
## Summary

The web dashboard is dark-only and hardcodes its entire palette as literal Tailwind utilities (`bg-zinc-950`, `text-zinc-400`, `emerald/amber/rose`) plus raw hex in the chart layer. There is no semantic color-token layer: the `@theme` block in `styles.css` defines only fonts. This issue introduces a design-token system, a first-class light theme, and themeable typography, so:

- operators can run the dashboard in light mode,
- a host page embedding the chromeless dashboard (embed mode, #716) can match the surface to its surroundings via the existing sanitized theme mechanism, which today supports exactly two CSS variables (`bg` / `fg`),
- the default dark render stays byte-for-byte identical throughout.

This is a large but highly mechanical migration. It is phased into independently landable PRs, each green on `pnpm typecheck && pnpm lint && pnpm test`.

## Current state (measured)

- `apps/web/src/styles.css` (~2,200 lines): `@theme inline` defines ONLY `--font-sans` / `--font-mono`. ~416 `@apply` lines carry literal palette utilities. A global base rule applies `border-zinc-800` to `*`. Roughly a dozen raw hex/rgba literals (`#09090b` embed shell bg, gauge/sparkline strokes `#34d399` / `#fbbf24` / `#fb7185`, highlight rgba, selection/scrollbar rgba).
- `.tsx` components: ~2,100 inline literal color-utility sites. Heaviest files: `pages/ProjectPage.tsx`, `components/project/GscSection.tsx`, `components/project/ActivitySection.tsx`, `pages/ReportPage.tsx`, `components/shared/AeroBar.tsx`, `components/project/BacklinksSection.tsx`.
- Charts: `components/shared/ChartPrimitives.tsx` exports hardcoded hex constants consumed as Recharts JS props: `CHART_TOOLTIP_STYLE`, `CHART_AXIS_TICK` (`#71717a`), `CHART_GRID_STROKE` / `CHART_AXIS_STROKE` (`#27272a`), `CHART_SERIES_COLORS` (8 hex), `PROVIDER_SERIES_COLORS` (per-engine), `CHART_NEUTRAL`, `CHART_TONE`. The tone hex is duplicated in CSS (`.gauge-fill-*`, `.sparkline-*`).
[...]

PR #768, opened 2026-07-02
feat(web): add design token foundation

arberx:
## Summary

Phase 1 for #767. Adds the web dashboard design-token foundation without migrating existing consumers yet.

- registers semantic color, tone, ring, and chart tokens in a static non-inline Tailwind `@theme` block
- exposes semantic utility aliases for the next migration slices
- adds compiler-level token assertions and SSR class-list baselines
- documents the new token rules in the root and web agent guidance
- bumps the root and publishable package versions to `4.104.1`

## Validation

- `corepack pnpm --filter @ainyc/canonry-web typecheck`
- `corepack pnpm --filter @ainyc/canonry-web test`
- `corepack pnpm --filter @ainyc/canonry-web lint` (0 errors, existing warnings)
- `corepack pnpm --filter @ainyc/canonry-web build` (existing large-chunk warning)
- commit hook `pnpm -r run lint` (0 errors, existing warnings)
- `git diff --check`

## Notes

Browser smoke was attempted against Vite on `localhost:4173`, but the in-app browser runtime failed before navigation with `sandboxCwd must be an absolute file URI`. The production web build verifies the Tailwind/Vite pipeline and the bundle contains the emitted chart tokens.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -11,9 +11,24 @@ The web dashboard follows a dark, professional analytics aesthetic inspired by *
 - Pages use a `page-header` (title + subtitle + optional actions) followed by sections separated by `page-section-divider`.
 
 ### Color & Theme
-- Background: `bg-zinc-950`. Cards/surfaces: `bg-zinc-900/30` with `border-zinc-800/60`.
+- Use semantic color tokens for new dashboard code. `apps/web/src/styles.css`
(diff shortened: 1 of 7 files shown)


arberx:
Superseded — Phase 1 (design-token foundation) landed on main via #771 (which included these commits). Closing.

PR #772, opened 2026-07-02
feat(discovery): seed hygiene - branded self-query filter, no-brand prompt rule, buyer anchoring

arberx:
## What

Three layers that stop discovery from spending probe budget on the customer's own brand:

1. **No-brand prompt rule**: `buildSeedPrompt` now names the customer's concrete brand identities and domains and forbids them in generated queries (buyers who already know the business by name are out of scope; the goal is queries where the customer must earn the mention).
2. **Deterministic filter backstop**: `filterBrandedSeedCandidates` (contracts) drops branded candidates that slip through, applied to the raw candidate list **before `seedCountRaw` is recorded**. The ordering is load-bearing: live sessions carried 37-60% brand share, and counting that mass in the denominator would deflate the retention ratio and false-trip collapse guards. Matching is conservative: whole-token phrase + squashed brand names ("AZ Coatings" also catches "azcoatings") and full canonical hosts, never the bare domain label, so a business at roofing.com does not lose every "roofing" query.
3. **Buyer anchoring**: optional `buyerDescription` on discover-run (API/CLI `--buyer`/MCP parity) adds a Buyer line + buyer-fit rule to the seed prompt, enforcing buyer-fit at generation time instead of post-probe.

Diagnostics: `discovery_sessions.seed_brand_filtered_count` (migration 89) records the per-session drop count.

## Why

On live local-vertical sessions, 25-46% of paid grounded probes were branded self-queries, and nearly every "cited" probe was the customer's own name, so the headline citation signal measured brand recall rather than buyer visibility, while the product's own taxonomy excludes branded-navigational clusters from ad groups (pure waste). Filtering also removes templated brand near-duplicates from the seed set, which raises retention and de-pressures the collapse guards.

## Notes

- Competitor brand names are untouched: only the customer's identities filter, so competitor comparatives survive.
- Version bumped 4.106.0 -> 4.107.0 in both package.json files.
- typecheck, 4962 tests, lint pass; new tests cover the filter's matching semantics (phrase/squashed/domain/whole-token/generic-label safety), the pre-count ordering with the persisted diagnostic, buyer pass-through to the seed dep, and the prompt rules.


arberx:
Both review findings validated and fixed in the follow-up commit:

- **Buyer now part of session identity.** Reproduced from the code path: the consolidation lookup keyed only on (project, icpDescription) while the buyer was forwarded after the reuse decision. `buyer_description` now persists on `discovery_sessions` (migration 90), joins the lookup null-safely (no-buyer only reuses no-buyer), and is serialized on the session DTO. Route tests cover: different buyer = fresh session, no-buyer vs with-buyer = fresh session, same buyer = consolidated, and row-level persistence.
- **Domain tokens normalize through `hostOf`.** Reproduced exactly as reported ('example.com reviews' kept with `canonicalDomains: ['https://www.example.com/path']`). Tokens now derive from the canonical host extractor with a raw-string fallback for unparseable values; regression test uses the raw-URL configuration.

PR #773, opened 2026-07-02
feat(web): bridge ChartPrimitives to CSS chart tokens (#767 phase 4)

arberx:
Phase 4 of #767 (stacked on #771). Wires the Recharts JS color constants to the `--chart-*` CSS tokens the foundation registered — charts become runtime-themeable while the default dark render stays byte-for-byte.

## What's here
- **`ChartPrimitives.tsx`:** every color constant is now `var(--chart-*, <original hex fallback>)` (tooltip, axis tick, grid/axis strokes, `CHART_SERIES_COLORS[1-8]`, `CHART_NEUTRAL`, `CHART_TONE`). `PROVIDER_SERIES_COLORS` stays **literal** — engine identity, not tone. Audited every consumer across 7 files: all pass the constants straight to SVG `fill`/`stroke` / inline style / `stopColor` (+ separate `stopOpacity`) — **zero string-math**, so `var()` strings are safe.
- **`styles.css`:** gauge/sparkline strokes move from the oklch tone-scale tokens onto the shared hex `--chart-tone-*` / `--chart-neutral-grid-line`, collapsing the CSS-side tone duplicates (gauges can't drift from charts) and restoring the exact original stroke hex. Drops the now-dead `--color-track`.
- **`test/chart-primitives.test.ts`:** asserts each constant bridges to its `--chart-*` token **and** that every JS fallback equals the CSS token default — closing the two-source-of-truth drift gap (it immediately caught an `rgba(…)` vs `rgb(… / …)` mismatch). Locks `PROVIDER_SERIES_COLORS` as fixed literal hex.

## Verification
Consumer audit clean (no slice/alpha-concat on the colors). Full web `test` (351), `typecheck`, `lint`, `build` green. The default render is unchanged: the CSS vars are always defined, so the hex fallbacks never activate at runtime; the gauge rewiring is hex→hex.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -88,7 +88,7 @@ The web dashboard follows a dark, professional analytics aesthetic inspired by *
 
 - Import chart components and shared constants from `components/shared/ChartPrimitives.js`.
 - Use `CHART_TOOLTIP_STYLE`, `CHART_AXIS_TICK`, `CHART_GRID_STROKE`, `CHART_AXIS_STROKE`, and `CHART_SERIES_COLORS` for consistent styling.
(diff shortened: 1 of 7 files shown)


arberx:
Superseded by #774 — same commit, rebased onto main after #771 landed. (This PR auto-closed when its base branch `arberx/issue-767-styles-migration` was deleted on merge.)

PR #775, opened 2026-07-02
feat(discovery): geo-aware probes, locations join session identity, downgrade-safety harness

arberx:
## What

1. **Geo-aware probes.** Discovery probes now carry the session's FIRST resolved service area into `executeTrackedQuery`, the exact mechanism sweeps use (provider renders "searching from City, Region, Country"). Location-free sessions probe byte-identically to today. Multi-location probe fan-out is deliberate future work; one geo context per session keeps probe cost flat.
2. **Locations join the consolidation identity** (the repo's identity-vs-tuning rule applied to an existing field): resolved locations change seed geo-constraints AND probe geo context, so same-ICP requests with different location subsets must never share a session. Locations persist on `discovery_sessions` (migration 91), join the reuse key, and ride the session DTO. Legacy NULL in-flight rows: reused on no-location projects (geo unambiguous), never reused on located projects (subset unknowable, conservative for the bounded 2h window).
3. **Standing downgrade-safety harness** (`packages/db/test/migration-downgrade-safety.test.ts`): platform tenants pin engine images and roll BACK by reprovisioning older images onto migrated volumes. The harness enforces, for every migration after the v88 baseline: additive-only statements, nullable-or-defaulted columns, an older binary (simulated via a new injectable version list on `migrate()`) boots a fully-migrated newer DB without throwing or re-applying, and old-writer INSERTs omitting new columns succeed. Future migrations are covered automatically.

## Why

Live local sessions probed 0/77 cited from a geo-less context, plausibly systematic downward bias on exactly the segment being sold. The identity gap was the same bug class as the buyer finding on the previous PR, caught this time by the documented audit instead of review. The downgrade harness makes rollback safety a tested invariant instead of a convention.

## Notes

- TDD throughout; 19 new tests, full suite 4983 green.
- Two CLI consolidation fixtures updated to seed sessions the way the route writes them since migration 91 (resolved locations persisted); the legacy-NULL semantics are pinned in dedicated route tests instead.
- No new request fields: `locations` already existed; MCP/CLI/OpenAPI unchanged except the additive session DTO field (client regenerated).
- Version bumped 4.107.0 -> 4.108.0 in both package.json files.


arberx:
Both findings validated and fixed:

- **P2 probe geo**: confirmed the divergence against job-runner's sweep resolution (`projectLocations.find(l => l.label === project.defaultLocation)`). The route now orders resolved locations default-first via a shared `orderLocationsDefaultFirst` contracts helper, so probes measure from the sweep's geo. Tests: default led to the front (callback + persisted session), no-op when the default is absent/unknown/already first, subset overrides that exclude the default unchanged.
- **P3 run() hooks**: confirmed the blind spot. Post-baseline migrations that define `run()` now fail the harness unless listed in an explicit `RUN_HOOK_ALLOWLIST` with a reviewed justification.

PR #815, opened 2026-07-14
fix(api): preserve tracked-query row identity across declarative replaces

arberx:
## What

`canonry apply`, `PUT /projects/:name/queries`, and `PUT /projects/:name/keywords` replaced a project's tracked queries with delete-all + reinsert, minting new row UUIDs even when the incoming list was textually identical to the stored basket. Because `query_snapshots.query_id` is ON DELETE SET NULL, every such write silently orphaned ALL of the project's historical snapshot links: FK-based readers (the analytics/metrics trend behind the dashboard's answer-engine chart) collapsed to "only one sweep" while text-fallback readers (`visibility-stats`, timeline) kept working. Confirmed in production: an apply that changed only `canonicalDomain` re-created 11 identical query rows and orphaned 1,567 snapshots.

All three routes now go through a shared `replaceProjectQueries` (`api-routes/src/query-replace.ts`):

- Diff by `normalizeQueryText` (trim + lowercase): unchanged texts KEEP their existing rows — id, createdAt, and provenance untouched.
- Casing/whitespace-only changes update the stored text in place on the kept row.
- Incoming duplicates (same normalized text) collapse to one row — a declarative list is a set; duplicate rows double-probe the same query on every sweep.
- Only genuinely removed rows are deleted, after the existing `preserveSnapshotQueryText` safety net (moved into the same module) stamps their text onto any referencing snapshot. The apply path never had this safety net before; now it does.

Migration **v98** (data-only `run()` hook, allowlisted with justification in the downgrade-safety test, per the v96 precedent) self-heals installs already hit: relinks FK-null snapshots to the same-project query with matching normalized text. Retired-query orphans have no match and correctly stay unlinked. Idempotent via the `query_id IS NULL` guard.

## Why

Root AGENTS.md "Idempotent writes": running apply twice with the same input must produce the same state — and row identity is part of that state, because those ids anchor the FK attribution for every historical sweep. Without this, any config apply (even one touching an unrelated field) resets the project's dashboard trend history.

## Tests

- 8 route-level regression cases (`api-routes/test/query-row-identity.test.ts`): identical applies preserve ids and createdAt; a canonicalDomain-only apply doesn't touch query rows; snapshots keep their FK across an apply; delta-only churn on a changed basket; casing-only in-place update; identical replace via `PUT /queries` and `PUT /keywords`; removed rows detach via the text safety net while kept rows stay linked.
- 4 migration cases (`db/test/relink-snapshot-query-ids.test.ts`): normalized-text relink; no cross-project links when two projects share a query text; retired/text-less orphans stay unlinked; existing FKs untouched + idempotent re-run.

Full gate: typecheck, lint, and the whole workspace suite (426 files / 5,180 tests) green. Version 4.117.0 → 4.117.1.


arberx:
All four findings addressed in the latest commit:

- **P1 (duplicate snapshots orphaned):** `replaceProjectQueries` now classifies same-normalized-text extras as `duplicates` and REPARENTS their snapshots onto the kept row (`UPDATE query_snapshots SET query_id = keptId`) before deleting the row — same normalized text is the same tracked query, so attribution survives the collapse. Regression tests cover the pair through both `PUT /queries` and `POST /apply`, asserting both rows' snapshots end on the kept id.
- **P1 (unique-constraint abort):** write ordering fixed — duplicates and removed rows are deleted before kept-row renames. The exact scenario (`Best AEO Agency` kept, `best aeo agency` also existing, incoming `best aeo agency`) is now a test and passes; after the dedup pass every kept row has a distinct normalized key and a rename target normalizes to its own key, so renames cannot collide with survivors or inserts (rationale in the module doc comment).
- **P1 (v98 vs Unicode):** the migration is now a TS `run()` hook that matches with the shared `normalizeQueryText` (packages/db already depends on contracts at runtime — same pattern as v96's classifier). New test: an `ÉCOLE DE COMMERCE PARIS` orphan relinks to the `école de commerce paris` query. The downgrade-safety allowlist entry stands (data-only UPDATEs, no schema change).
- **P2 (preview reports old behavior):** the diff logic is extracted into a pure exported `diffProjectQueries`, consumed by BOTH `replaceProjectQueries` and the replace-preview route — one diff, two consumers, no drift. Preview now reports normalized-text `added/removed/unchanged` and a `snapshotImpact` restricted to genuinely removed rows (kept rows keep their ids; duplicates reparent). For `alpha,beta → alpha,gamma` with 3+2 snapshots it now reports `removed: ['beta']`, `affectedQueries: 1`, `snapshotsDetached: 2`. Tests updated, plus a casing-only no-op case. CLI/MCP dry-run guidance is fixed transitively since they render this endpoint's response.

Full gate green: typecheck, lint, 426 test files / 5,193 tests.

PR #880, opened 2026-07-31
fix(provider-claude): require web search under a persisted measurement contract

arberx:
Rebased onto #879. Behaviour and persistence now land together.

## What

Claude tracked queries run under a named, persisted measurement contract. Retrieval is required through an API control, and both what happened and the policy it happened under are written to every snapshot.

```
retrieval_status:   used | not-used | unknown | not-applicable
retrieval_contract: native-auto-v1 | search-required-v1
```

`search-required-v1` = unmodified user query, no system prompt, `tool_choice` pinned to `web_search`.

## Why

Claude decides for itself whether a query warrants a search, and Sonnet 5 decides far more conservatively than Sonnet 4.6. Measured over 60 days:

| Model | Searched | Total | Rate |
|---|---|---|---|
| claude-sonnet-4-6 | 136 | 136 | 100% |
| claude-sonnet-5 | 84 | 119 | 71% |

All 35 unsearched answers stored zero cited domains and zero mentions, indistinguishable at rest from an answer that searched and did not mention the brand, while sitting in the denominator of every visibility rate.

| Project | Claude rows | Searched | As reported | Searched only |
|---|---|---|---|---|
| azcoatings | 80 | 45 | 10.0% | 17.8% |
| gjelina-hotel | 39 | 39 | 23.1% | 23.1% |

Gjelina is unaffected: its queries name an entity or place, which still triggers search. Claude-specific — `openai chat-latest` searched on all 129 of its rows, so its zero-cited rows are legitimate measurement.

## Why `tool_choice`, not a system prompt

A system prompt is probabilistic and steers persona, tone, and source policy as well as retrieval, contaminating the substance being measured. `tool_choice` is the documented control for requiring a tool invocation. `max_uses` is a ceiling and cannot raise the floor.

Measured on the five queries that lost retrieval in production (`claude-sonnet-5`, n=5 per mode):

| Mode | Retrieval | Avg cited domains | Avg answer chars |
|---|---|---|---|
| Native (`auto`) | 1/5 | 1.4 | 2,920 |
| System prompt | 5/5 | 8.0 | 5,155 |
| **`search-required-v1`** | **5/5** | **5.0** | **4,520** |

Forcing does not suppress the answer we parse: it prefills the assistant turn so no preamble emits before the tool call, but every response returned `end_turn` with a complete answer.

Keeps `web_search_20250305`; the `_20260209+` variants add post-search dynamic filtering and do not make search more likely.

## The invariant

**A search-policy change must never produce an unmarked snapshot.**

Both fields are **required** on `RawQueryResult`, not optional. The type system now forces every adapter to state a value rather than silently omit one, so a new provider cannot be added unmarked. Adding them failed the build until all six adapters declared a position.

| Provider | Status | Reason |
|---|---|---|
| claude | real detection | the only marker demonstrated to discriminate: 35 stored negatives vs 84 positives |
| local | `not-applicable` | plain chat completion, no web access |
| openai | `unknown` | detection deferred |
| gemini | `unknown` | detection deferred |
| perplexity | `unknown` | detection deferred |
| cdp | `unknown` | scraped rendered UI, no reliable signal |

The three deferrals are deliberate. Each candidate marker (`web_search_call`, `groundingMetadata`, `search_results`) is present on **100%** of its provider's rows over 30 days, so none has been shown to discriminate a non-retrieving answer. Wiring one up would hardcode `used` while looking like a measurement. Detection lands per provider once each marker is validated against a known negative.

`retrieval_status` is orthogonal to #879's `captureStatus` and is not derived from it. An extraction that ran and found zero sources is legitimately `complete` and says nothing about whether a search happened.

## Historical rows

Not backfilled. They were produced under provider-native behaviour, but writing that in would launder an assumption into an observation. Null means "predates the field" and must not be read as `not-used`.

Migration 112 verified against a 1 GB copy of production: applied in 24ms, both columns present, all 4,007 existing rows null, integrity check clean.

## Tests

Full suite 6,145 passing across 507 files. Typecheck clean. Generated API client regenerated, `gen:check` clean, nullable preserved on both enums.

Each behavioural test verified by ablation rather than observed green:

1. `tool_choice` is set and no system prompt is sent
2. a `server_tool_use` `web_search` block records retrieval
3. searched-but-uncited stays distinct from unsearched, asserted identical on cited domains and citation state and separated only by retrieval status
4. retrieval is read from the search call, not the recovered query text
5. an undeterminable response is `unknown`, never `not-used`
6. retrieval survives the adapter boundary
7. both JobRunner insert branches persist the contract, proven by removing only the plain branch and watching only its tests fail
8. pre-migration rows read back null

## Comparability

Claude answers produced under this contract are not method-comparable to those produced before it. Month-over-month reporting spanning the cutover should say so, and `retrieval_contract` now makes the cutover queryable rather than a thing you have to remember.


arberx (commented):
Good production investigation. I think the implementation should first make the measurement contract explicit.

The API cannot exactly reproduce Claude.ai: the consumer product's system instructions, routing, and search policy are not public. Also, `web_search_20260209+` adds **post-search dynamic filtering**; it does not make search more likely. [Anthropic web-search versions](https://platform.claude.com/docs/en/agents-and-tools/tool-use/web-search-tool)

I would avoid the `"consumer-facing assistant"` system prompt. It is probabilistic and can change persona, tone, source policy, and answer substance—not only retrieval.

There are two defensible modes:

- **Native Claude behavior:** raw user query, `tool_choice: auto`, and record whether Claude searched.
- **Search-required benchmark:** retain `web_search_20250305` and use the API control `tool_choice: { type: "tool", name: "web_search" }`, then validate final-answer extraction. Anthropic documents `tool_choice` as the way to require a tool invocation. [Tool-use overview](https://platform.claude.com/docs/en/agents-and-tools/tool-use/overview)

The latter approximates a search-grounded chat but should not be described as an exact Claude.ai reproduction.

Whichever contract we choose should be named/versioned on the run or snapshot so trends cannot silently mix methods. Retrieval rate, unconditional citation/mention rates, and citation-given-retrieval are separate metrics.

Tests should verify the behavioral boundary:

1. `server_tool_use(name="web_search")` records retrieval.
2. Searched-but-uncited stays distinct from unsearched.
3. Required mode rejects or marks a response with no search call.

The current “searched and cited nothing” fixture contains a citation, so it does not prove that distinction.

arberx:
Don't take the conflict against #879 now. Let #879 land, then rebase #880 and fold the persistence into #880 before #880 merges.

The invariant I care about is: **no search-policy behavior change should produce unmarked snapshots**. Provider-local fields that are dropped at the adapter boundary do not satisfy that, because the contract cannot be reconstructed later.

I would put both fields on the snapshot, since retrieval semantics differ by provider:

- `retrievalStatus: used | not-used | unknown | not-applicable`
- `retrievalContract: native-auto-v1 | search-required-v1`

One distinction: I would not redefine #879's `captureStatus: complete`. “Citation extraction completed and found zero sources” can legitimately be complete; it says nothing about whether retrieval ran. Retrieval status should remain a separate signal.

So: no avoidable conflict today, but make #880 dependent on #879 and land the forced-search behavior plus its persisted contract atomically.

PR #884, opened 2026-07-31
feat(web): add Target portfolio workflow

arberx:
## Overview\n\nAdds the reviewable Target portfolio workflow on top of the measurement adapters.\n\n- Adds the non-embed Portfolio route and a five-step sitemap-to-report workflow.\n- Supports 213-scale Target review, durable URL-coverage decisions, reusable query sets/templates, Target-level context, and reporting-only groups.\n- Validates and diffs drafts before CAS-bound immutable publication; publishing never starts a run.\n- Reads exact-revision reports with honest incomplete/no-result states and URL evidence drill-down.\n- Uses generated SDK wiring and strips Portfolio/unknown tabs from embed mode.\n- Contains synthetic data only; no customer fixtures, names, domains, or slugs.\n\n## Stack\n\n#881 → #896 → #906 → #884. #879 is unaffected.\n\n## Validation\n\n- 528 test files / 6,347 tests\n- Typecheck: 29/29 workspaces\n- Lint: 0 errors\n- Full workspace build\n- Generated client drift check\n- Plugin sync check at v4.141.0\n- Diff check and customer-data scan\n- Isolated browser QA through publish and exact-revision empty report; no provider run\n

PR #911, opened 2026-08-02
fix(web): land a published plan on its report, not on setup

arberx:
Two founder-reported issues from walking the live instance. Stacked on #884, which owns this surface.

## What was wrong

Opening a project that already had **Active revision 1** showed an empty "Import sitemap" form. A configured project looked unconfigured, and the badge saying a revision was active alongside a blank setup form read as though nothing had saved.

The path field also arrived prefilled with `/locations/{slug}` — an example belonging to a different site, presented as a value rather than a placeholder, so it read as "this is your rule".

## Change

| State | Lands on |
|---|---|
| Published plan, no unsaved draft | **Report** |
| Unsaved local draft | where the draft was (unchanged) |
| No plan at all | Import, a genuine first run (unchanged) |

Setup steps stay reachable from the published state; this is a landing change, not a lockout. The path example moves from the field's value to its placeholder.

Note the root cause was partly a race: the landing stage was computed once at mount, before the plan query resolved, so a project with a published plan was classified as having none. Both the initial state and the seeding effect now agree.

## Validation

- `apps/web` 68 files / 534 tests, typecheck clean
- 4 new tests in `portfolio-landing.test.tsx`, **proven by reverting**: with the fix backed out, "lands a published plan on its report" and "never prefills the path field" both fail. The landing assertion checks the report is actually rendered rather than merely that Import is absent, because the old behaviour landed on Targets and would have passed a negative-only assertion.

### DIFF
--- a/apps/web/src/components/project/PortfolioSection.tsx
+++ b/apps/web/src/components/project/PortfolioSection.tsx
@@ -374,10 +374,18 @@ export function PortfolioSection(props: PortfolioSectionProps) {
   })
   const [draft, setDraftState] = useState<PortfolioSetupDraft | null>(initialState.draft)
   const [draftSource, setDraftSource] = useState<'empty' | 'active' | 'local'>(initialState.source)
-  const [stage, setStage] = useState<Stage>(() => initialState.draft ? 1 : 0)
+  // Published plan already in hand at mount goes straight to the report; an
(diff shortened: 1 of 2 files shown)


arberx:
Superseded by #912. This branch's history was rewritten; the same changes are in #912 as an ordered commit series on top of main.

PR #944, opened 2026-08-05
fix(ga): fold truncated landing paths into the page they belong to

arberx:
Stacked on #943. Review that one first; this diff is only the second commit.

## What

One page's sessions arrive split across several spellings of its own path, because a URL that was line-wrapped in a plain-text context or pasted out of chat output reaches analytics cut short. On a production table:

| Path | Sessions |
|---|---|
| `/aeo-methodology` | 42 |
| `/aeo-methodolo` | 7 |
| `/aeo-met` | 9 |
| `/aeo-meth` | 5 |

The real page was under-reported by 33%.

## The rule

Nothing in the string tells a truncation from a genuine 404, so `resolveTruncatedPaths` resolves only against the pages the most recent site audit actually crawled, and only when the answer is unambiguous:

- the observed path is not itself a known page — a real page keeps its identity even when it prefixes others;
- it is a **strict** prefix of exactly one known page;
- both have the same segment depth, so `/blog` never folds into `/blog/a-post`.

A project that has never been audited has no page list, which makes this a no-op rather than a guess.

Run against the real table — 267 observed paths against 39 audited pages — it produces exactly the three folds above and no false positives.

## Performance

The obvious form compares every observed path against every known one. At a few thousand of each (the largest project on the instance has 3,164 distinct landing paths) that measured **~220ms per request**, which a dashboard read should not pay. The lookup now binary-searches to the sorted run of pages sharing the prefix: **~2.5ms**, same output.

## Behaviour changes

`/ga/traffic` applies its top-N limit after the fold rather than in SQL. A fragment and the page absorbing it can sit either side of the cut, and folding into a page the limit had already dropped would lose those sessions outright. The grouped read is therefore unbounded — bounded in practice by distinct landing pages, and `/ga/coverage` already read that way.

Merged rows sum `sessions` and `organicSessions`/`directSessions`. `users` takes the max instead: it is a COUNT DISTINCT at the grain GA was asked for, so adding it across fragments would count one visitor once per fragment.

## Tests

- `packages/contracts/test/resolve-truncated-paths.test.ts` — 14 cases: the four-way real-world split, the section-root refusal, ambiguity, the sentinel, an empty page list, case sensitivity, and the run boundaries the binary search introduces (match at the first and last element, and a lexicographic neighbour that does not share the prefix).
- `packages/api-routes/test/ga-truncated-path-fold.test.ts` — 6 route-level cases including the no-audit no-op and `limit=1`, which is the case that fails if the limit ever moves back into SQL.

Making the fold a no-op fails 4 of those 6:

```
× folds the truncated spellings once an audit says which pages exist
    AssertionError: expected 42 to be 63
× conserves total sessions across the fold
    AssertionError: expected [ …(7) ] to have a length of 4 but got 7
× folds before applying the limit, so a fragment below the cut is not lost
    AssertionError: expected 42 to be 63
× applies the same fold to /ga/coverage
    AssertionError: expected 42 to be 63
```

The other two are guards for cases where it must not act, and pass either way by design.

`api-routes` + `contracts`: 199 files, 3462 tests, all passing.

## Caveat

This is a heuristic. The three conditions make it conservative — it declines rather than guesses — but a genuine 404 that happens to be a unique same-depth prefix of exactly one real page would fold. It runs at read time and rewrites no stored rows, so it is reversible by reverting.


arberx:
Updated after self-review.

Moving the top-N limit out of SQL is required when a fold can happen — a fragment and the page absorbing it can sit either side of the cut. But the original version dropped the SQL limit *unconditionally*, including for projects that have never been audited, where there is no page list and no fold is possible. Those pulled every grouped row to return 50 of them. The largest project on the instance has **3,164 distinct landing paths**, so that was 3,164 rows fetched per dashboard load to accomplish nothing.

The page list is now read first and decides how the query is bounded:

- **page list present** — query comes back whole, fold, then `slice(0, limit)`
- **page list empty** — nothing can fold, so SQL applies the limit as it did before

Test added: `still honours the limit when there is no page list to fold against`.

PR #947, opened 2026-08-05
fix(bing): retry throttled Bing calls instead of failing them

arberx:
Stacked on #946. Review that first; this diff is only the second commit.

## What

`integration-bing` was the only integration with **no retry at all**. GBP, Places, GA4 and every LLM provider already go through `withRetry`; Bing failed on the first attempt.

Bing throttles per host and per API key, and the host limit is shared by every project on an instance — so a refresh inspecting a few dozen URLs hits it no matter how careful one project is. The live evidence, from the runs table:

| Time | Succeeded | Failed |
|---|---|---|
| 14:28 | 10 | 20 |
| 13:40 | 10 | 23 |
| 02:53 | 10 | 47 |

Exactly ten succeed, then everything behind them fails. 229 throttle failures over three days.

## Changes

**`BingApiError` carries the condition.** It now holds Bing's own `ErrorCode` and an `isThrottle` flag (4 = per-key, 5 = per-host, plus HTTP 429 which is rate limiting by definition). Bing reports throttling as a 400 with the detail in the body, so the status alone cannot tell a throttle from a bad request. Carrying the code makes it structural rather than something callers parse out of prose, and lets #946's predicate see it.

**Every call goes through `withRetry`.** `bingFetchOnce` does the request; the exported `bingFetch` wraps it. Retry is a property of the client, not something each call site remembers.

Policy: 4 retries, 2s base, 30s ceiling, honouring `Retry-After`. The base is well above the 1s default on purpose — the observed pattern is a burst succeeding and everything behind it throttling within the same second, which a sub-second backoff walks straight back into.

## What still fails fast

Auth, validation, not-found, and non-throttle Bing codes fail on the first attempt, with a test pinning it. This matters: the same instance logged **550 `NotAuthorized` errors** earlier this year, and retrying each of those five times would have quintupled the load against a wall.

A non-JSON error body (an HTML gateway page) parses to no code and is treated as non-throttle, also tested.

## Behaviour change

HTTP 429 is now retried rather than thrown immediately, so a persistent 429 surfaces after the retry budget instead of at once. The existing `throws BingApiError on 429` test is **updated to assert that** — it now checks the error still surfaces, carries `isThrottle: true`, and that exactly 5 attempts were made. It was not deleted or skipped.

## Tests

9 new in `bing-throttle-retry.test.ts` (fake timers, so backoff costs no wall-clock), covering: retry-then-succeed on both throttle codes, exhausting the budget, and the four things that must NOT retry. `integration-bing` is 22 tests, all passing.

## Note on the dependency

`integration-bing` had no `dependencies` at all; this adds `@ainyc/canonry-contracts: workspace:*`, matching how `integration-google-analytics` declares it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "canonry",
   "private": true,
-  "version": "4.148.3",
+  "version": "4.148.4",
   "type": "module",
   "packageManager": "pnpm@10.28.2",
   "scripts": {
--- a/packages/canonry/package.json
+++ b/packages/canonry/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@canonry/canonry",
-  "version": "4.148.3",
+  "version": "4.148.4",
   "type": "module",
   "description": "Agent-first open-source AEO operating platform - track how answer engines cite your domain",
   "license": "FSL-1.1-ALv2",
--- a/packages/integration-bing/package.json
+++ b/packages/integration-bing/package.json
@@ -16,5 +16,7 @@
     "test": "vitest run --config ../../vitest.package.config.ts",
     "lint": "eslint src/ test/"
   },
-  "dependencies": {}
+  "dependencies": {
+    "@ainyc/canonry-contracts": "workspace:*"
+  }
 }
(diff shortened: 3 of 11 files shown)


arberx:
Updated after self-review — the original version of this PR had a real bug.

Wrapping retry at the transport layer meant *every* call retried, including three POSTs: `addSite`, `submitUrl`, and `submitUrlBatch`. The last two spend a daily quota (`BING_SUBMIT_URL_DAILY_LIMIT`, 10,000). A 5xx or dropped connection after Bing had already accepted a submission would replay it and burn quota that cannot be reclaimed. The PR body claimed retry was safe because it is a property of the client; that is true for reads and I overstated it for writes.

Now split by what the failure tells us:

| Failure | GET | POST |
|---|---|---|
| Throttle | retry | retry |
| 5xx / network | retry | **fail fast** |

A throttle is rejected at the gate, so a repeat cannot double-submit — that is the case this PR exists for and it still applies everywhere. A 5xx or network error is ambiguous about whether the server acted, so it is only replayed where a repeat is harmless.

Two tests added. Removing the guard fails `does NOT retry a write that failed ambiguously` with `expected 5 to be 1` — five submissions where one was intended.

arberx:
Auto-closed by GitHub when its base branch `ax/generic-rate-limit-retry` was deleted on merge of #946 — my mistake for passing `--delete-branch` on a stacked PR. Cannot be reopened once the base is gone.

Continued in #949: same branch, same commit, retargeted to `main`.

PR #957, opened 2026-08-05
feat(gsc): derive index coverage from impressions so it scales past the quota

arberx:
Stacked on #956. Review that first; this diff is only the second commit.

## The constraint

Coverage came only from the URL Inspection API — ~7.1s and one unit of a **2000/property/day** quota per page.

| Pages | Full sweep | Daily quota |
|---|---|---|
| 290 (demand-iq) | 34 min | 15% |
| 1,000 | 2.0 hrs | 50% |
| 5,000 | 9.9 hrs | **250% — impossible** |

In practice the cap bit far earlier: the inline pass took the top 50 by clicks, so a 290-page site reported coverage from 50 of them and said nothing about the other 240.

## The free signal

**A page with at least one impression in the window IS in Google's index** — Google cannot have shown it otherwise. That arrives in rows the sync already fetches, in under a second, for every page, at any site size. (This is standard practice, not a novel trick — deriving indexation from search-analytics data rather than per-URL inspection is the documented workaround for the quota.)

The old selection was backwards: it spent the scarce resource on *top pages by clicks*, i.e. precisely the pages guaranteed to have impressions.

## The trap this is built around

The inverse does **not** hold. No impressions is not evidence of exclusion — a page can be indexed and simply never rank in the window. Treating silence as "not indexed" would manufacture bad news at exactly the moment coverage stopped being truncated.

So there are **three** states, and an unmeasured page is `unknown`, not a zero.

Precedence per page in `deriveIndexCoverage`:

| | Signal | Result |
|---|---|---|
| 1 | impressions > 0 | `indexed` — certain, free. Checked first, so it also overrides a stale not-indexed verdict from before a page was fixed |
| 2 | an inspection verdict | `indexed` / `not-indexed` per `indexingState` |
| 3 | neither | `unknown`. Never `not-indexed` |

Pages inspected but absent from the window are included too — dropping them would silently shrink the denominator and flatter the number.

## Schema

`gsc_coverage_snapshots` gains `unknown_pages` plus provenance: `verified_by_inspection` and `derived_from_impressions`, so a later change can say how much of the figure is a real Google verdict. Named `unknown_pages` because bare `unknown` is a reserved SQL keyword.

Migration 125 creates the table if missing before altering it — `ALTER TABLE` on a missing table raises an error the runner does not swallow (only duplicate-column), and the legacy-schema path hits exactly that.

## Behaviour change — say this out loud

**Coverage numbers will move.** They now span the whole property instead of up to 50 inspected pages, and `unknown` is a bucket that did not exist. A site whose coverage looked complete because only its best 50 pages were counted will now show unknowns.

That is the truncation becoming visible, not a regression. Please don't revert it in a panic.

## Tests

9 cases on the helper: impression as proof, `unknown` never collapsing to `not-indexed`, stale-verdict override, inspected-but-unranked pages, per-query row summing, bucket conservation, a 5000-page site resolved with zero inspections, empty input. Plus an end-to-end case asserting the sync writes coverage for every page while inspecting nothing.

Collapsing `unknown` into `notIndexed` fails 4 of them:

```
× reports a page with no impressions and no inspection as unknown, NOT not-indexed
    expected 'not-indexed' to be 'unknown'
× conserves every page across the three buckets
    expected [ 1, 2, +0 ] to deeply equal [ 1, 1, 1 ]
× scales to a site far past the inspection quota without inspecting anything
× writes coverage for every page from impressions, inspecting nothing
```

3,135 tests pass across `db` / `contracts` / `canonry`. Docs updated: `data-model.md` for the columns, `contracts/AGENTS.md` for the utility.


arberx:
**Correction — do not merge this without #959.**

This PR claims coverage now spans the whole property. Self-review found that was **false in production**.

`inspect-sitemap` also writes the same `(project, date)` row with a delete-then-insert, and `server.ts` chains it off a successful `gsc-sync`. It still counted inspections alone and never set `unknown_pages` / `verified_by_inspection` / `derived_from_impressions`, so they fell back to `DEFAULT 0`. Every derived figure this PR adds was overwritten minutes after being written, on the exact path #956 relies on.

#959 fixes it by routing both writers through one `writeCoverageSnapshot` that reads both signals from the database, so neither can hold a different view of the same facts.

The miss is worth recording: both writers were individually correct and individually tested. Nothing exercised them **in sequence**, which is the only way they run.

PR #959, opened 2026-08-05
fix(gsc): give the coverage snapshot a single writer

arberx:
Stacked on #958. **This fixes a defect in #957 that self-review found — #957 should not merge without it.**

## What was broken

`gsc-sync` and `inspect-sitemap` both write the same `(project, date)` coverage row with a delete-then-insert, and `server.ts` chains the second off the success of the first. After #957 made `gsc-sync` derive coverage across the whole property, the two disagreed:

```
gsc-sync completes
  └─ writes 3-state coverage   indexed=2  unknown=1  derived=2   ✅
       ↓ chained via maybeRefreshGscCoverage
inspect-sitemap runs minutes later
  └─ DELETEs the row
  └─ writes 2-state coverage   indexed=2  unknown=0  derived=0   ❌
```

`inspect-sitemap` still counted inspections alone and never set the three new columns, so they fell back to `DEFAULT 0`. **Whole-property coverage silently degraded to inspection-only coverage every cycle** — and #957's claim that coverage now spans the property was false in production, on the exact path #956 depends on.

## The fix

Both paths call one `writeCoverageSnapshot`, which reads **both** signals from the database rather than taking them from the caller.

That framing is the point. Passing the data in would have fixed this instance; reading it means neither writer can hold a different view of the same facts, so the class is gone. `inspect-sitemap` needs impressions it never fetches, which is what forced the read.

Net: `gsc-sync` −54 lines, `inspect-sitemap` −53, one 112-line shared writer.

## Why it was missed

Each writer was individually correct and individually tested. Nothing exercised them **in sequence**, which is the only way they run.

`gsc-coverage-single-writer.test.ts` runs sync-then-sitemap against a real SQLite database and asserts the derived figures and provenance survive the rewrite. Reverting the helper to inspection-only fails 2 of its 4 cases:

```
× a chained inspect-sitemap does not erase what gsc-sync derived
    expected +0 to be 2
× keeps derived coverage when inspections exist for only some pages
    expected +0 to be 2
```

## One test moved — flagging explicitly

The coverage assertions in `gsc-sync-no-inspection.test.ts` drove a hand-rolled db mock. Now that impressions are read back out of `gsc_search_data`, that mock cannot represent them, so the **numeric** assertions moved to the real-database test — a better home regardless, since it can run both writers against each other.

The mock test keeps its shape check and its no-inspection guarantee. Nothing was weakened to make it pass; the failure told me the fixture was the wrong tool, so I moved the assertion rather than lowering it.

## Side effect

The 1-hour `COVERAGE_REFRESH_MIN_INTERVAL_MS` throttle used to mean a manual refresh inside that window did nothing for coverage at all (since #956 removed gsc-sync's own inspection). `gsc-sync` now writes derived coverage on every run regardless of the throttle, so that gap closes too.

3,145 tests pass across `canonry` / `contracts` / `db`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "canonry",
   "private": true,
-  "version": "4.148.8",
+  "version": "4.148.9",
   "type": "module",
   "packageManager": "pnpm@10.28.2",
   "scripts": {
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "canonry",
   "private": true,
-  "version": "4.148.9",
+  "version": "4.148.10",
   "type": "module",
   "packageManager": "pnpm@10.28.2",
   "scripts": {
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "canonry",
   "private": true,
-  "version": "4.148.10",
+  "version": "4.148.11",
   "type": "module",
   "packageManager": "pnpm@10.28.2",
   "scripts": {
--- a/packages/canonry/package.json
+++ b/packages/canonry/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@canonry/canonry",
-  "version": "4.148.9",
+  "version": "4.148.10",
   "type": "module",
   "description": "Agent-first open-source AEO operating platform - track how answer engines cite your domain",
   "license": "FSL-1.1-ALv2",
--- a/packages/canonry/package.json
+++ b/packages/canonry/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@canonry/canonry",
-  "version": "4.148.8",
+  "version": "4.148.9",
   "type": "module",
   "description": "Agent-first open-source AEO operating platform - track how answer engines cite your domain",
   "license": "FSL-1.1-ALv2",
--- a/packages/canonry/package.json
+++ b/packages/canonry/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@canonry/canonry",
-  "version": "4.148.10",
+  "version": "4.148.11",
   "type": "module",
   "description": "Agent-first open-source AEO operating platform - track how answer engines cite your domain",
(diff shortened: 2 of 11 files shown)


arberx:
Two more commits on this branch, both from smoke-testing against a copy of the production database.

**`d8c98ce7` — pages are matched on normalized path.** The three sources spell the same page differently, and matching them raw was double-counting a domain migration:

```
sitemap     https://canonry.ai/privacy    no trailing slash
GSC page    https://canonry.ai/           trailing slash
inspection  http://ainyc.ai/              pre-migration scheme AND host
```

On real data that inflated canonry's coverage denominator from **50 pages to 85** — 35 phantom entries that were the same pages under the old `ainyc.ai` host. No unit test caught it because every fixture used consistent URLs; only a real database has a domain migration in its history.

It also seeds the page universe from the latest site audit. Without that the `unknown` state was unreachable: GSC only returns rows for pages that *had* an impression, so a crawled page with no impressions and no inspection was absent from the denominator entirely rather than counted as unknown.

**`5dfb5731` — a sweep is capped at 1500 URLs and says so.** `inspect-sitemap` was uncapped against a 2000/property/day quota. Past it the run did not fail cleanly — it ground for hours, exhausted the allowance partway, and tripped the circuit breaker, so the operator saw a *failure* when the budget had simply run out. It now logs `sitemap.over-budget` in plain language and completes as `partial`, because pages were left without a verdict.

**Verified on production data** (a `.backup` copy — the live database was never touched and remains on schema 124): migration applies in 21ms with all eight table row counts identical, coverage computes in 20-68ms per project, and demand-iq's denominator widens from 170 to 205 pages.

One clarification on `unknown` reading 0 across all four live projects: that is correct, not dead. `inspect-sitemap` has already inspected every crawled page on these sites. It fires for a newly published page, an aborted inspection, or a large site whose budget has not reached everything — which is exactly the 1500-cap case above.

279 files / 3149 tests pass.

PR #963, opened 2026-08-06
fix(bing): move the sweep out of the browser

arberx:
The Bing sweep ran as a loop **in the dashboard**, issuing one HTTP call per URL from the tab. Every failure traced to that.

## The symptoms, and the one cause

| Symptom | Cause |
|---|---|
| "if I leave the page the refresh stops" | the loop is client JS guarded by `signal.aborted` at 6 points, tied to component lifecycle |
| deploying a fix changed nothing | the browser served a **cached bundle** — the tab, not the server, decided the rate |
| "10 Bing inspection(s) failed" every time | 10 concurrent against a per-**host** throttle Bing reports as `400 ErrorCode 5` |
| it got **slower** after tuning 10→3 | 76 URLs at 3/batch = **26 batches** ≥1s apart, against 8 before |
| no CLI/MCP equivalent | batching logic existed only in the dashboard — the UI/CLI parity rule forbids this |

The user diagnosed it: *"if I leave the page, it seems like the refresh stops."*

## Retry is not the problem

#949 works. The deployed engine logs **50 `http.throttled`, 40 `http.retry`**, and failed runs take 10–24s of backoff instead of failing instantly. It cannot compensate for a client firing bursts at a host-limited API — synchronised callers back off and retry together and keep the throttle asserted.

Tuning the burst size in the UI was treating the symptom. Three rounds of it.

## The fix is deletion

`bing-inspect-sitemap` **already exists** and **already paces at ~1 req/sec** server-side. The UI even called it — but only when there were zero prior inspections.

So the per-URL loop is replaced by the same trigger-and-poll shape `syncGoogle` uses. No new machinery.

- the run survives closing the tab
- pacing lives in one place instead of being guessed at in the browser
- a stale bundle can no longer change the rate
- the CLI and MCP get the capability the dashboard was hoarding

A `partial` run reports through `notices` rather than `failures` — pages left unverified is not a failure, and saying so wrongly is what made this look broken.

`inspectBingUrl` stays; it still backs the single-URL manual inspection.

## Follow-up, deliberately not here

`bing-inspect-sitemap` paces with a bare `setTimeout` rather than `inspectUrlsPaced`, so it has no jitter, no circuit breaker, and no sweep-size cap. Moving it onto the shared helper would give it the same protection the GSC sweep has. Recorded in `~/bing-server-side-sweep-scope.md`.

## Verify after deploy

**Hard-refresh the tab** — a cached bundle silently keeps the old behaviour, which cost an hour of confusion here with the server correct and the browser wrong.

```sql
SELECT substr(started_at,12,5) min, SUM(status='completed') ok, SUM(status='failed') failed
FROM runs WHERE kind LIKE 'bing-inspect%' AND started_at > '<deploy>'
GROUP BY substr(started_at,1,16) ORDER BY min DESC;
```

Success = the 10-ok/10-failed signature is gone, and the run keeps progressing after navigating away.

82 files / 727 web tests pass.

### DIFF
--- a/apps/web/src/pages/ProjectPage.tsx
+++ b/apps/web/src/pages/ProjectPage.tsx
@@ -871,41 +871,48 @@ function SearchConsoleSection({
         notices.push('Google sync is still running — search data will appear shortly')
       }
 
-      // --- Bing: re-inspect previously known URLs, or fall back to sitemap ---
-      // Bing throttles per HOST, shared by every project on the instance, and
-      // reports it as a 400 rather than a 429. Ten simultaneous inspections
-      // reliably tripped it: the first batch landed, everything behind it was
-      // rejected. The engine now retries with backoff, but ten synchronised
-      // callers back off and retry together — a thundering herd that keeps the
-      // throttle asserted. Fewer in flight, spaced apart, gets more work done
-      // than more in flight being rejected.
-      const BING_CONCURRENCY = 3
-      const BING_BATCH_SPACING_MS = 1000
+      // --- Bing: trigger the server-side sweep and poll, like syncGoogle ---
+      //
+      // This used to loop here, issuing one HTTP call per URL from the browser.
+      // Every failure mode traced to that: the sweep died when you navigated
+      // away (it is client JS guarded by `signal.aborted`), a cached bundle
+      // silently kept the old batch size, and the burst size was tuned in the
+      // UI against a limit enforced on the server. Bing throttles per HOST,
+      // shared by every project on the instance, so a browser is the wrong
+      // place to decide the rate.
+      //
+      // `bing-inspect-sitemap` already walks the sitemap server-side at ~1
+      // req/sec. Triggering it means the run survives closing the tab, the
+      // pacing lives in one place, and the CLI/MCP get the same capability —
+      // the UI/CLI parity rule this loop was violating.
       async function syncBing() {
         if (!bingConnection?.connected) return
-        const inspections = await queryClient.fetchQuery({
-          ...getApiV1ProjectsByNameBingInspectionsOptions({ client: heyClient, path: { name: projectName } }),
-          staleTime: 0,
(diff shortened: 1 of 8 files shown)


arberx:
**Follow-up completed in this PR** (commit `81707ce7`) — the one the description deferred.

`bing-inspect-sitemap` paced with a bare `setTimeout(1000)`: correct spacing, nothing else. No jitter, no circuit breaker, no size cap. A wholly throttled property ground through the whole sitemap against a wall, spending the daily allowance to produce nothing and reporting it as a pile of per-URL errors.

It now uses `inspectUrlsPaced` — the same driver as the GSC sweep.

**Two changes were needed to share it, and the first is load-bearing:**

`isRetryable` is now injectable (defaulting to the GSC predicate). That predicate decides **two** things — what gets retried AND what advances the circuit breaker. Bing reports a throttle as `400` with `ErrorCode 5`, which the GSC predicate does not recognise, so left hardcoded **Bing's breaker would never trip** and every throttle would look like a permanent per-URL failure. Bing passes `isRetryableHttpError`, which understands the condition semantically since #946.

`inspectUrlsPaced` is now generic over the result type. It was bound to `GscUrlInspectionResult` while Bing's `getUrlInfo` returns `BingUrlInfo` — but the driver never inspects the result, only hands it to `onResult`, so the binding was incidental.

Also applies the same `INSPECT_SWEEP_MAX_URLS` cap and `sitemap.over-budget` warning the GSC sweep got.

**Concurrency pinned to 1 for Bing.** The throttle is per-HOST and shared by every project on the instance, and the observed pattern is that a burst lands and everything behind it is rejected. Most conservative setting until measured; the rate gate is unchanged either way.

**Version 4.150.1** — main moved to 4.150.0 while this branch was open, so the earlier 4.148.13 was a downgrade. The plugin-manifest test caught it.

194 files / 1989 canonry tests pass, plus the 82/727 web suite from the first commit.

PR #1046, opened 2026-08-25
feat(google-ads): performance endpoint, CLI, and metrics UI

arberx:
## Context

Google Ads syncs store a `campaign-metrics` snapshot on every run, but nothing read it back. The only Google-marketing surface was the Conversion Integrity checklist, so an operator who connected an account and synced saw no spend, clicks, impressions, or conversions anywhere.

## Changes

- `GET /projects/:name/google-ads/performance` reads the latest stored metrics + inventory snapshots and returns computed totals, a densified daily series, per-campaign totals, and a prior-equal-period comparison. Zero provider calls: this serves stored data only, so it never spends against the ad account.
- `canonry google-ads performance <project> [--window 7d|14d|28d]` for parity. `--format json` emits the DTO verbatim.
- `canonry_google_ads_performance` MCP tool (read tier).
- A metrics section on the project conversions tab: KPI tiles, a daily chart, and a campaign table, matching the density of the Search Console section.
- Conversion Integrity setup collapses the two settled connections into one status line and promotes the conversion declaration, which is the only recurring action. No action was removed.

## Decisions worth review

**Windows are 7d/14d/28d, not 30d/90d.** A snapshot holds 31 days, at most 30 closed. A prior-equal-period comparison needs 2N closed days, so 28d returns `comparison: null` with `insufficient-history` rather than a comparison the stored data cannot support. Deeper history needs the sync to accumulate rather than overwrite; that is a separate change and this DTO accommodates it by changing only what the route reads from.

**The capture day is excluded.** Snapshots are taken mid-day, so the newest day is partial and including it renders a fabricated decline on the right edge of every chart. `asOfDate` is the newest closed day and every window ends there. The cutoff derives from the payload's own `fetchedAt`, never a server clock, so a stale snapshot cannot slide it.

**Zero denominators are null, never 0.** An undefined 0/0 is not a measured zero, and rendering it as 0 invents a fact. This holds through the DTO, the UI, and the CLI.

**Two different caps, reported separately.** `truncated` is the provider row cap. The 50-campaign query cap is a distinct limit: an account above it is summed from the queried subset while `truncated` stays false. Gating the caveat on `truncated` alone would print a subtotal as an account total, so coverage is reported from `campaignsQueried` vs `campaignsInInventory` independently.

## Verification

`pnpm verify` green: 704 test files, 9,196 tests.

New tests assert exact math rather than shape: daily series sums to totals; `ctr` unrounded; zero-denominator cases null; fractional conversions preserved; the open day absent from totals, the series, and per-campaign figures; calendar gaps densified as measured zeros; a campaign missing from inventory kept with `name: null`; 28d comparison unavailable; the no-snapshot empty shape returning 200 rather than 404; and two reads at different wall-clock instants returning byte-identical payloads.


packages/api-routes/src/google-marketing.ts
```diff
@@ -1410,6 +1796,57 @@ export async function googleMarketingRoutes(app: FastifyInstance, opts: GoogleMa
...
+    const candidateRun = queuedRun(projectId, RunKinds[kind])
+    let run: typeof runs.$inferSelect | typeof candidateRun = candidateRun
+    let queued = false
+    app.db.transaction((tx) => {
+      const existing = tx.select().from(runs).where(and(
+        eq(runs.projectId, projectId),
+        eq(runs.kind, RunKinds[kind]),
+        or(eq(runs.status, RunStatuses.queued), eq(runs.status, RunStatuses.running)),
+      )).orderBy(desc(runs.createdAt)).get()
+      if (existing) {
```
arberx: **[P1] Reselection can lose its replacement sync.** Selection increments the generation and clears snapshot pointers, but this reuses an older queued or running run. That executor then fails its generation CAS and no successor remains queued. Deduplicate per generation or always queue a successor.

packages/canonry/src/google-marketing-runtime.ts
```diff
@@ -927,7 +959,24 @@ class DefaultGoogleMarketingRuntime implements GoogleMarketingRuntime {
...
+    // removed ones included, so the daily query is scoped to where the money
+    // went. A failure here is not fatal; selection degrades to inventory order.
+    const rankingWindow = defaultMetricsWindow(this.#now(), customerResult.data.timeZone)
+    const rankedCampaignIds = await client
+      .getCampaignSpendRanking(customerId, {
+        ...rankingWindow,
+        limit: GOOGLE_ADS_CAMPAIGN_METRICS_MAX_CAMPAIGNS,
+      })
+      .then(result => result.data.map(row => String(row.campaign.id)))
+      .catch(() => undefined)
```
arberx: **[P1] Removed-campaign spend can still be silently omitted.** Ranking failures are swallowed, and exactly 50 rows are fetched without a 51st-row sentinel. Inventory fallback excludes removed campaigns, allowing historical spend and conversions to be omitted with `truncated: false`. Fail the sync or propagate incompleteness and fetch limit plus one.

packages/api-routes/src/google-marketing.ts
```diff
@@ -1448,6 +1885,7 @@ export async function googleMarketingRoutes(app: FastifyInstance, opts: GoogleMa
         diff: { loginCustomerId: body.loginCustomerId ?? null, customerId: body.customerId },
       }))
     })
+    queueSelectionSync(request, project.id, 'google-ads-sync', opts.onGoogleAdsSyncRequested)
```
arberx, line 1888: **[P1] Auto-sync completion is unobservable to callers.** Selection queues work but returns only status, so the dashboard cannot track the run. Its immediate refetch occurs before completion and no terminal invalidation refreshes status, performance, or conversion options. Return and track the queued run.

arberx (commented):
Reviewed the current head across API/runtime, CLI/MCP, and the dashboard. CI and focused tests pass, but the correctness and operator-state issues below should be resolved before merge.

packages/canonry/src/cli-commands/google-marketing.ts
```diff
@@ -54,6 +56,27 @@ function snapshotLimit(
...
+ */
+function performanceWindow(
+  input: Parameters<CliCommandSpec['run']>[0],
+  usage: string,
+): GoogleAdsMetricsWindow | undefined {
+  const raw = getString(input.values, 'window')
+  if (raw === undefined) return undefined
+  const parsed = googleAdsMetricsWindowSchema.safeParse(raw)
+  if (!parsed.success) {
+    throw usageError(`Error: --window must be one of 7d, 14d, 28d\nUsage: ${usage}`, {
```
arberx: **[P2] The CLI advertises an invalid window.** The contract accepts `7d`, `14d`, and `30d`, while this error and usage advertise `28d`. Consequently `28d` fails and valid `30d` is undiscoverable.

apps/web/src/components/project/ConversionIntegrityWorkspace.tsx
```diff
@@ -439,6 +537,12 @@ export function ConversionIntegrityWorkspace({ projectId, projectName }: { proje
     ...getApiV1ProjectsByNameConversionTrackingContractsOptions({ client: heyClient, path: { name: projectName } }),
     staleTime: GOOGLE_MARKETING_STALE_MS,
   })
+  const conversionOptionsQuery = useQuery({
```
arberx, line 564: **[P2] Option-read failures look like setup work.** The query error state is discarded. Failed reads and successful empty syncs both become manual inputs saying to run another sync. Render distinct error-with-retry, unsynced, and synced-empty states.

apps/web/src/components/project/GoogleAdsPerformanceSection.tsx
```diff
@@ -0,0 +1,536 @@
...
+    { key: 'ctr', label: 'CTR', value: formatGoogleAdsRatio(totals.ctr), seriesKey: null },
+    { key: 'cpc', label: 'CPC', value: formatGoogleAdsMicros(totals.cpcMicros, currency), seriesKey: null },
+  ]
+
+  const SERIES_META: Record<SeriesKey, { label: string; color: string; axisId: string; formatValue: (v: number) => string }> = {
+    // Micros stay integers all the way to the axis; the formatter is the single
+    // render edge that turns them into money.
+    costMicros: { label: 'Spend', color: CHART_SERIES_COLORS[1]!, axisId: 'spend', formatValue: (v) => formatGoogleAdsMicros(v, currency) },
+    conversions: { label: 'Conversions', color: CHART_SERIES_COLORS[0]!, axisId: 'conversions', formatValue: (v) => CONVERSION_FORMAT.format(v) },
+    clicks: { label: 'Clicks', color: CHART_SERIES_COLORS[2]!, axisId: 'conversions', formatValue: (v) => COUNT_FORMAT.format(v) },
```
arberx: **[P2] Clicks flatten conversions.** These metrics have different units and magnitudes but share one axis. The fixture has 913 clicks versus 37.5 conversions, reducing conversions to roughly 4% of the scale. Give Clicks a distinct axis.

apps/web/src/components/project/GoogleAdsPerformanceSection.tsx
```diff
@@ -0,0 +1,536 @@
...
+  if (micros === null || !Number.isFinite(micros)) return GOOGLE_ADS_NOT_AVAILABLE
+  // No resolved account currency: show the magnitude without asserting a unit.
+  // Printing "$" on a EUR account is a wrong number, not a missing one.
+  if (currency === null) return AMOUNT_FORMAT.format(micros / 1_000_000)
+  return formatMicros(micros, currency)
+}
+
+/** A raw ratio (0.0731) as a display percentage. Never rounded before this point. */
+export function formatGoogleAdsRatio(ratio: number | null): string {
+  if (ratio === null || !Number.isFinite(ratio)) return GOOGLE_ADS_NOT_AVAILABLE
```
arberx, line 86: **[P2] Small positive rates render as zero.** One decimal place converts a measured `0.04%` rate into `0%`. Preserve the nonzero signal with a `<0.1%` display.

apps/web/src/components/project/ConversionIntegritySection.tsx
```diff
@@ -547,44 +567,43 @@ function ConversionIntegritySetup({
...
     {
-      title: 'Google Ads account',
+      title: 'Google Ads',
+      state: workspace.googleAds.state,
       ready: googleAdsReady,
       summary: workspace.googleAds.selection,
-      detail: 'Connect read-only access and choose the customer account.',
       changeLabel: 'Change Google Ads account',
       onChange: onChangeGoogleAdsSelection,
       action: connectionUnavailable ? retryAction : googleAdsSetupAction(workspace),
```
arberx: **[P2] Retry appears beside the healthy provider.** Any provider failure assigns the retry action only to the Google Ads row and removes it from Tag Manager. When only GTM fails, Retry appears beside healthy Google Ads. Attach it to the failed row or make it global.

apps/web/src/components/project/ConversionIntegrityWorkspace.tsx
```diff
@@ -408,6 +430,82 @@ function DisconnectConnection({
...
+          {options.map((option) => (
+            <option key={option.id} value={option.id}>
+              {option.name}
+              {option.detail ? ` (${option.detail})` : ''}
+              {option.active ? '' : ' — inactive'}
+            </option>
+          ))}
+        </select>
+      )}
+      <button
```
arberx, line 521: **[P2] This label contains multiple interactive controls.** A button and select or input are nested inside one label, producing invalid and ambiguous activation semantics. Separate the label, form control, and mode-switch button.

packages/canonry/src/commands/google-marketing.ts
```diff
@@ -233,6 +237,85 @@ export async function googleAdsSync(
...
+  }
+  // The row cap and the 50-campaign query cap are DIFFERENT limits. An account
+  // with more than 50 campaigns is summed from the queried subset while
+  // `truncated` stays false, so reporting coverage only on `truncated` would
+  // print a subtotal as the account total. Compare the counts directly.
+  if (result.source.campaignsInInventory > 0 && result.source.campaignsQueried < result.source.campaignsInInventory) {
+    console.log(`Coverage:     ${result.source.campaignsQueried} of ${result.source.campaignsInInventory} campaigns; totals are a subset of the account.`)
+  }
+  if (result.source.truncated) {
+    console.log(`Note:         the provider row cap was hit, so these totals are a subset sum.`)
```
arberx: **[P2] This reports the wrong truncation cause.** Runtime combines provider-row truncation with the 50-campaign selection cap, but every case is attributed to the provider row cap here. Report the actual completeness reason.

PR #1050, opened 2026-08-25
feat(traffic): per-day content-crawl counts on the daily series

arberx:
The daily series carried only `crawlerHits`, the full count including robots, sitemaps and assets. The totals object has carried `crawlerContentHits` for a while, so a chart could show the honest number for a window but not per day. Anything drawing a daily "pages crawled" line had to either use the inflated total or invent the split client-side.

## Change

`crawlerContentHits` is now on `trafficSeriesPointSchema`.

Path classification is per path, not per hit, so the series query groups by bucket **and** normalized path and runs the same `segmentCrawlerHits` the totals already use. Distinct normalized paths per window is small, so the extra grouping key is cheap.

`crawlerHits` keeps its full-count contract and is unchanged.

## Test

The new test asserts the two figures **disagree** on the existing mixed-path fixture: 3 content against 15 total, with sitemap, robots, asset and other making up the rest. Asserting either is merely non-zero would pass on a source that never fetches a sitemap, which is exactly the case this field does not need to exist for.

It also pins the series against the totals on the same request, and checks every point carries the key including zero-filled days.

One existing exact-shape assertion was updated to include the new field. That test doing its job is why the wire shape cannot drift silently here.

## Parity

The CLI already prints `crawlerContentHits` from totals. The series reaches it through `traffic events --format json`, which carries the new field automatically.

## Verification

`pnpm verify` exit 0: 703 test files, 9,196 tests.

No version bump: under the 100-line threshold.

### DIFF
--- a/apps/web/src/components/project/ActivitySection.tsx
+++ b/apps/web/src/components/project/ActivitySection.tsx
@@ -1074,7 +1074,13 @@ export function ClickThroughActivity({ projectName, window: windowProp }: {
               </Card>
             )}
 
-            <div className="grid gap-4 xl:grid-cols-[minmax(0,0.9fr)_minmax(0,1.5fr)]">
+            {/* Stacked, not side by side. The breakdown is a six-column table
+                (source, medium, channel, sessions, share, users) and at ~60% of
+                the row it wrapped or truncated the source names, which are the
+                column a reader actually scans. Full width also matches the
+                approved mock. The summary tiles above it are already a 3-up
+                grid, so they fill the width without changes. */}
+            <div className="grid gap-4">
               <Card className="surface-card p-5">
                 <div className="mb-4">
                   <p className="eyebrow eyebrow-soft">Summary</p>
--- a/apps/web/src/components/project/ActivitySection.tsx
+++ b/apps/web/src/components/project/ActivitySection.tsx
@@ -28,6 +28,7 @@ import {
   urlSearchText,
   useClientTable,
 } from '../shared/DataTableControls.js'
+import { AiTrafficHistoryPanel } from './AiTrafficHistoryPanel.js'
 import { InfoTooltip } from '../shared/InfoTooltip.js'
 import { ToneBadge } from '../shared/ToneBadge.js'
 import type { MetricsWindow } from '@ainyc/canonry-contracts'
@@ -64,6 +65,17 @@ import {
 import { buildAiChartData } from '../../lib/ai-chart-helpers.js'
 
 const TRAFFIC_WINDOWS: MetricsWindow[] = ['7d', '30d', '90d', 'all']
+
+// The server-side lane is queried in minutes, the GA lane by window name. One
+// control drives both, so the mapping lives here rather than in either consumer.
+// 'all' is capped at 90 days: `sinceMinutes` has no unbounded representation,
+// and an uncapped scan is not what the operator is asking for from a chart.
+const TRAFFIC_WINDOW_MINUTES: Record<MetricsWindow, number> = {
+  '7d': 7 * 24 * 60,
+  '30d': 30 * 24 * 60,
+  '90d': 90 * 24 * 60,
+  all: 90 * 24 * 60,
+}
 
 const SOURCE_COLORS = CHART_SERIES_COLORS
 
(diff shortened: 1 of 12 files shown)


apps/web/src/components/project/ActivitySection.tsx
```diff
@@ -65,6 +66,17 @@ import { buildAiChartData } from '../../lib/ai-chart-helpers.js'
...
 
+// The server-side lane is queried in minutes, the GA lane by window name. One
+// control drives both, so the mapping lives here rather than in either consumer.
+// 'all' is capped at 90 days: `sinceMinutes` has no unbounded representation,
+// and an uncapped scan is not what the operator is asking for from a chart.
+const TRAFFIC_WINDOW_MINUTES: Record<MetricsWindow, number> = {
+  '7d': 7 * 24 * 60,
+  '30d': 30 * 24 * 60,
+  '90d': 90 * 24 * 60,
+  all: 90 * 24 * 60,
```
arberx, line 84: **[P1] `All` represents different windows.** GA omits the window for `All`, while server traffic maps it to 90 days. One control therefore labels two incomparable ranges identically. Give both lanes the same bounds or label them separately.

packages/api-routes/src/traffic.ts
```diff
@@ -4479,17 +4479,31 @@ export async function trafficRoutes(app: FastifyInstance, opts: TrafficRoutesOpt
...
       const crawlerSeries = app.db
         .select({
           bucket: crawlerSeriesBucket,
+          pathNormalized: crawlerEventsHourly.pathNormalized,
           hits: sql<number>`COALESCE(SUM(${crawlerEventsHourly.hits}), 0)`,
         })
         .from(crawlerEventsHourly)
         .where(crawlerWhere)
-        .groupBy(crawlerSeriesBucket)
+        .groupBy(crawlerSeriesBucket, crawlerEventsHourly.pathNormalized)
```
arberx: **[P1] This series query has unbounded bucket-by-path cardinality.** It materializes every group before the later event limit applies, so a large site can produce millions of rows for a 90-day page load. Aggregate by a persisted or SQL-derived path class and add a high-cardinality test.

apps/web/src/components/project/AiTrafficHistoryPanel.tsx
```diff
@@ -0,0 +1,182 @@
...
+      fetches: acc.fetches + p.aiUserFetchHits,
+      visits: acc.visits + p.aiReferralLandedHits,
+    }),
+    { crawlers: 0, fetches: 0, visits: 0 },
+  ), [points])
+
+  if (events.isLoading) {
+    return <Card className="surface-card p-5"><div className="text-sm text-muted">Loading AI traffic history…</div></Card>
+  }
+  if (points.length === 0) {
```
arberx: **[P2] Error and empty states are inverted.** A request error yields no points and is presented as no activity. Successful empty windows are densified into zero points, so this branch is unreachable for the state it claims. Handle `isError` explicitly and detect an all-zero series.

arberx (commented):
Reviewed the current head. CI and focused tests pass, but the correctness, scaling, state, and accessibility issues below should be resolved before merge.

apps/web/src/components/project/ActivitySection.tsx
```diff
@@ -215,18 +227,40 @@ function ServerActivityPanel({ projectName }: { projectName: string }) {
...
       <ServerActivityPanel projectName={projectName} />
-      <ClickThroughActivity projectName={projectName} />
+      <AiTrafficHistoryPanel
+        projectName={projectName}
+        sinceMinutes={TRAFFIC_WINDOW_MINUTES[trafficWindow]}
+      />
+      <ClickThroughActivity
+        projectName={projectName}
+        window={trafficWindow}
+        onWindowChange={setTrafficWindow}
```
arberx: **[P2] The traffic range control disappears without GA4.** This wrapper owns the shared state, but the only picker remains inside `ClickThroughActivity`, which returns before rendering it when GA4 is disconnected. The server chart remains visible and locked to 30 days. Move the picker into `ActivitySection`.

apps/web/src/components/project/AiTrafficHistoryPanel.tsx
```diff
@@ -0,0 +1,182 @@
...
+          value={compact(totals.visits)}
+          caption="arrived from an AI answer"
+          tooltip="Visits your server answered that came from an AI engine. Requests answered with a redirect are excluded: a redirect is a hop, not an arrival."
+        />
+      </div>
+
+      <div className="mb-2 flex items-center justify-between gap-3">
+        <h3 className="text-sm font-semibold text-heading flex items-center gap-1.5">
+          Machines reading your site
+          <InfoTooltip text="Crawlers index pages for later. Page fetches happen while an engine is answering someone. Both are machines, not people." />
```
arberx, line 340: **[P2] `InfoTooltip` renders an interactive button inside the heading.** That alters the accessible heading name and violates the project heading rule. Make the heading and tooltip siblings; the second chart heading has the same issue.

PR #1168, opened 2026-09-16
Multi-property UI review fixes, and the agency report back in the app

arberx:
Answers the Loom review of the multi-property UI, and restores the agency report inside the app so the two report surfaces match again.

**Portfolio UI, from the review**
- Project identity moves to the breadcrumb and a context row replaces the page header; the measurement scope control rides along and works across tabs.
- North-star numbers lead the AI Visibility tab as separate tiles, each with its change inline and the compared sweep named once in the caption.
- Filters collapse into a toolbar with removable tokens; markets link to their groups.
- The five collapsed rows that end the tab share one row and one panel spacing.

**The agency report is back in the app**
- `ReportPage.tsx` renders all 19 agency sections again (#432 hid the tab and #558 deleted the components, so every agency table lived only in the download). It opens on Client, with an Agency toggle; embeds stay client-only.
- Section copy, order and display helpers move into `packages/contracts/src/report-sections.ts`, which both renderers now read. The HTML output is byte-identical.

**Parity guards, because copy in two renderers drifts**
- Byte snapshots for 2 audiences x 6 fixtures, plus section-outline goldens written only from the HTML renderer, so a SPA-only change cannot make parity pass.
- A content golden records tile values, table rows, badges with tone, list rows and link targets. It caught a real defect: both surfaces badged "Gained citation" in the alarm tone under Wins.

**Phone fixes**
- The Simple overview, the Advanced toolbar and every table in the downloadable report stay inside the viewport at 375px and 320px; the report's insights table keeps every column readable.

**Checks**
- web 2,026, api-routes 3,757, contracts 1,391, CLI 3,053; four typechecks, lint, `build:web`, and the codegen, plugin and Val skill drift checks.
- Verified on the demo at 1280px and 375px for a Simple and an Advanced project, both audiences.
- 5.4.0: `main` and npm both reached 5.3.0 while this was in flight.


apps/web/src/components/project/VisibilityTrendSection.tsx
```diff
@@ -266,30 +452,166 @@ function selectedReportPopulation(report: VisibilityReportResponse, queryKey?: s
...
+  const filtersButton = useRef<HTMLButtonElement>(null)
+  const controlId = useId()
+  const panelId = `${controlId}-filters`
+  const { filterOptions, selection: served } = report
+  const measurement = served.measurement
+  const population = selectedReportPopulation(report, selection.queryKey, selection.answer)
+  // A clean URL asks for every class until normalization records the served one.
+  const queryClass = selection.queryClass === 'all' ? population.queryClass : selection.queryClass
+  const tokens = visibilityFilterTokens(selection, report)
+  const propertyLink = served.mode === 'advanced' && served.scope.kind === 'property' ? renderPropertyLink?.({ id: served.scope.id, label: served.scope.label }) : null
```
arberx: **[suggestion]** The Advanced toolbar’s Property details link is the one control that reads the served report (`served.scope`) instead of the URL selection. `useVisibilityReportFirstPage` keeps the previous report as placeholder data while a new property (or a leave-property) request loads, so the link’s `id`/`label` and the `aria-label` (`Property details for ${label}`) still name the previous property. Tokens and query type already bind to the URL; this link can navigate to the wrong property during that window.

**Suggestion:** Derive the link from `selection.measurementScope` / `measurementScopeKey` (label from `selectedScopeOption(report.scopeOptions, selection)` or the URL key). Keep choice lists on the displayed report, but do not let placeholder `served.scope` drive navigation.

apps/web/src/pages/ProjectPage.tsx
```diff
@@ -2650,64 +2691,137 @@ function ProjectPageContent({
...
-                        : 'Run AI sweep'}
-              </WriteButton>
-            </div>
-          ))}
+      ) : (
+        // The topbar breadcrumb names the project. The row keeps the page's one
+        // h1 for assistive tech and narrow screens, except on Report, which
+        // renders its own.
+        <div className="project-context-row">
+          {tab !== 'report' ? (
```
arberx: **[suggestion]** The context row skips its `h1` on the Report tab so `ReportPage` can own the title, but `ReportPage` only renders that `h1` after a successful load. While the report query is loading or has failed, the tab has no heading at all (previously the project page header always supplied one). Embeds are fine because they keep the in-page header.

**Suggestion:** Keep a visually hidden project `h1` on the report tab until `ReportPage` mounts its own, or render the report header (eyebrow/title) in the loading and error states as well.

PR #1169, opened 2026-09-16
fix(cli): clarify plugin lockstep tip and print serve URL last

arberx:
## Summary

`cnry serve` printed a plugin-cache mismatch as `Claude Code v4.129.0 does not match the running Canonry v5.3.1`, which read as a CLI upgrade notice even when the runtime was already current. The localhost URL was also buried between scheduler boot logs and catch-up doctor HTTP.

- Name the Canonry plugin, not the host app: `the Canonry plugin for Claude Code (v4.129.0) does not match this Canonry v5.3.1. Update or reinstall \`canonry@canonry\` in Claude Code.`
- Wait for doctor catch-up before reporting serve ready, so the URL is the last startup line.
- Do not stack that plugin tip on top of `UPDATE_AVAILABLE`. They are different channels (npm vs Claude/Codex plugin cache). If the CLI itself is behind, say that once; after the runtime is current, the plugin tip is the leftover lockstep check. Doctor still reports plugin mismatch.

## Test plan

- [x] `pnpm exec vitest run --project contracts packages/contracts/test/skills.test.ts`
- [x] `pnpm exec vitest run --project api-routes packages/api-routes/test/doctor-agent.test.ts`
- [x] `pnpm exec vitest run --project canonry packages/canonry/test/skills-install.test.ts packages/canonry/test/scheduler.test.ts packages/canonry/test/system-jsonl-degrade.test.ts packages/canonry/test/server-listen-lifecycle.test.ts`
- [x] `pnpm check` and typecheck of contracts, api-routes, canonry
- [ ] After merge, `cnry serve` on a host with a stale Claude Code plugin and a current CLI should print the URL last, then the plugin tip (not a CLI upgrade notice)
- [ ] `cnry serve` with a cached `UPDATE_AVAILABLE` should not also print the plugin tip


packages/canonry/src/commands/serve.ts
```diff
@@ -122,14 +123,18 @@ export async function serveCommand(format: CliFormat = 'text'): Promise<void> {
...
+      // catch-up logs, so the URL is not buried in the middle of boot output.
       console.log(`\nCanonry server running at ${url}`)
       console.log(`Open ${url}/setup to map your site and run your first Page Health scan.`)
       if (shouldWarnAboutRemoteSetup(host)) {
         console.log('First-run dashboard password setup is unauthenticated only on loopback; complete setup from this machine first or use a bearer cnry_... key.')
       }
       console.log('Press Ctrl+C to stop.\n')
       const nudge = getMissingUserSkillsNudge(process.env.HOME, getAgentPluginState())
-      if (nudge) process.stderr.write(`${nudge.message}\n`)
+      if (shouldPrintServeSkillsNudge(nudge, readCachedUpdateAvailable())) {
```
arberx: **[bug]** `shouldPrintServeSkillsNudge` is supposed to avoid stacking a plugin-lockstep tip on top of an `UPDATE_AVAILABLE` notice that already printed. The CLI prints that notice from `readCachedUpdateAvailable()` before dispatch, then fires `checkLatestVersionForCli()` in the background; that write can land during `createServer` / listen / doctor catch-up. This call then re-reads the cache. If the cache was empty or still equal to the running version at process start (TTL expired, first fetch, or last check failed without a version), no upgrade notice was shown, but a completed npm fetch now makes `updateAvailable` truthy and the plugin tip is dropped. The new catch-up wait makes that race likely instead of theoretical.

**Suggestion:** Gate on the snapshot the CLI already used (or a process-local “notice printed” flag), not a second disk read after startup work. The skip should mean “we already told them about a CLI upgrade,” not “the cache became newer while we waited.”

packages/canonry/src/server.ts
```diff
@@ -3646,8 +3646,16 @@ export async function createServer(opts: {
       for (const run of opts.db.select({ id: researchRuns.id, projectId: researchRuns.projectId }).from(researchRuns).where(eq(researchRuns.status, ResearchRunStatuses.queued)).all()) {
         dispatchResearchRun(run.id, run.projectId);
       }
-      runtimeStartupSettled = true;
-      resolveRuntimeStartup();
+      void scheduler.waitForStartupCatchUp().then(
```
arberx: **[suggestion]** `waitForServerRuntimeStartup` now waits for every catch-up `onDoctorRequested` promise. That callback is the full scheduled pass: sequential live doctor checks (GSC/GA/GBP/Bing/WordPress token refresh and listing, 30s Google timeouts) plus `notifier.onHealthChecked`, which awaits webhook send with 10s timeouts and 1s/4s/16s retries. Serve therefore delays the URL, machine-format `{started:true}`, and SIGINT/SIGTERM handler install until that work finishes. A first-observation warn (including plugin mismatch) can fire `health.degraded` and hold the banner on webhook retries. The comment on `waitForStartupCatchUp` excludes long jobs so they do not delay the banner; doctor was treated as cheap HTTP logs, which it is not.

**Suggestion:** Keep runtime-startup as “scheduler started,” install signal handlers then, and wait only for `runDoctor` (or a short timeout) before printing the URL. Do not await webhook delivery or unbounded third-party checks on the ready path.

packages/canonry/src/commands/skills.ts
```diff
@@ -498,12 +501,31 @@ export function emitInstallSummary(summary: SkillsInstallSummary, format?: strin
...
+   */
+  source: 'plugin' | 'legacy-skills'
   /** Skills missing from `~/.claude/skills/`. */
   missing: BundledSkillName[]
   /** Skills already installed there. */
   installed: BundledSkillName[]
 }
 
+/**
+ * Serve prints at most one version story. A plugin-lockstep tip on top of
```
arberx: **[suggestion]** The `shouldPrintServeSkillsNudge` JSDoc narrates the product history (“at most one version story,” leftover channel after the runtime is current) instead of stating the rule. The implementation also skips every `source: 'plugin'` nudge, including unverified-cache copy, which is not a version story.

**Suggestion:** Replace the JSDoc with the predicate (skip plugin-source nudges when an upgrade notice is already showing). If unverified cache should still print beside `UPDATE_AVAILABLE`, split `source` or key off mismatch vs unverified.

packages/contracts/src/skills.ts
```diff
@@ -22,6 +22,42 @@ export interface AgentPluginState {
...
+function joinPluginSubjects(subjects: string[]): string {
+  if (subjects.length <= 1) return subjects[0] ?? ''
+  if (subjects.length === 2) return `${subjects[0]} and ${subjects[1]}`
+  return `${subjects.slice(0, -1).join(', ')}, and ${subjects.at(-1)}`
+}
+
+/**
+ * Plugin cache version vs running Canonry runtime. Names the plugin, not the
+ * host app, so "Claude Code v4.129.0" is never read as the Claude Code version.
+ */
```
arberx: **[suggestion]** The mismatch formatter comment embeds the incident (“so Claude Code v4.129.0 is never read as the Claude Code version”) rather than a constraint the next reader needs.

**Suggestion:** Keep a one-liner that the string names the plugin cache version, not the host app version.

packages/canonry/src/commands/skills.ts
```diff
@@ -533,20 +555,23 @@ export function getMissingUserSkillsNudge(
...
-        .map((client) => `${client === 'claude-code' ? 'Claude Code' : 'Codex'} v${agentPlugin.verifiedClientVersions?.[client] ?? 'unknown'}`)
-        .join(', ')
-      problems.push(`${versions} ${mismatchedClients.length === 1 ? 'does' : 'do'} not match the running Canonry v${PACKAGE_VERSION}.`)
+      problems.push(`${formatCanonryPluginVersionMismatch({
+        mismatchedClients,
+        verifiedClientVersions: agentPlugin.verifiedClientVersions,
+        runningVersion: PACKAGE_VERSION,
+      })}.`)
     }
+    const hintClients = unverifiedClients.length > 0 ? unverifiedClients : mismatchedClients
```
arberx: **[nit]** When both unverified and mismatched clients exist, the update hint is built only from unverified clients, so a mismatched client is described in the tip and then omitted from “Update or reinstall … in …”.

**Suggestion:** Pass the union of both lists into `formatCanonryPluginUpdateHint`.

